PR #2, opened 2022-04-18
Admission control threat model.

### LINKED ISSUE #129: Re-check security posture against the k8s admission control threat model
## Acceptance criteria

Following the k8s admission controller threat model document listed in [kubernetes/sig-security/issues/7](https://togithub.com/kubernetes/sig-security/issues/7), check all threats and mitigations to see if we need to re-implement or add anything. Check if we want to provide a production setting with specific policies deployed by default (pod-privileged not allowed on same node as controller, etc).

In addition, list any stack differences if any on the Kubewarden stack, and provide a threat model analysis for those. 

Create followup cards with the work to be done if necessary.

jvanz:
Adds a RFC to discuss the admission control threat model and its mitigations.

Fix https://github.com/kubewarden/kubewarden-controller/issues/129

### DIFF
--- a/rfc/0006-threat-model.md
+++ b/rfc/0006-threat-model.md
@@ -1,0 +1,254 @@
+|              |                                  |
+| :----------- | :------------------------------- |
+| Feature Name | Kubernetes admission control threat model                           |
+| Start Date   | 04/18/2022                          |
+| Category     | Security                       |
+| RFC PR       | https://github.com/kubewarden/rfc/pull/2  |
+| State        | **ACCEPTED**                     |
+
+# Summary
+[summary]: #summary
+
+Describe the threat model for the admissions control, how it can effect Kubewarden
+and the possible mitigations.
+
+# Motivation
+[motivation]: #motivation
+
+The goal of this is turn Kubewarden more secure to operate.
+
+## Examples / User Stories
+[examples]: #examples
+
+As Kubewarden developer, I want to know which are the weak point of the Kubernetes admission control feature, so I can
+mitigate this problem.
+
+As Kubewarden user, I want to know which are the weak points of the Kubernetes admission control, so I can apply the mitigation
+in my cluster.
+
+# Detailed design
+[design]: #detailed-design
+
+The threat model of the Kubernetes admission control list 18 possible threats.
+All of them will be discussed sharing a brief description of it, possible mitigation
+and what we, as Kubewarden team, should do about it (if necessary). As well as
+help our users to be aware of the risks.
+
+If more details about each threat is needed, go to the original document here:
+https://github.com/kubernetes/sig-security/tree/main/sig-security-docs/papers/admission-control
+
+
+## Threat 1 - Attacker floods webhook with traffic preventing its operation
+
+An attacker who has access to the Webhook endpoint, at the network level, could
+send large  quantities of traffic, causing an effective denial of service to the
+admission controller.
+
+**Mitigation**
+
+Webhook fails closed. In other words, if the webhook does not respond in time,
+for any reasion, API server should reject the request. We are safe on this.
+Kubewarden default behavior  already does that.
+
+## Threat 2 - Attacker passes workloads which require complex  processing causing timeouts
+
+An attacker, who can access the admission controller at a network level, passes
+requests to  the admission controller which require complex processing, causing
+timeouts as the  admission controller uses compute power to process the workloads
+
+**Mitigation**
+Webhook fail closed and authenticate callers. We are safe, Kubewarden already does that.
+
+## Threat 3 - Attacker exploits misconfiguration of webhook to bypass
+An attacker, who has rights to create workloads in the cluster, is able to exploit
+a mis-configuration to bypass the intended security control
+
+**Mitigation**
+Regular reviews of webhook configuration catch issues.
+
+
+### Threat 4 - Attacker has rights to delete or modify the k8s webhook  object
+
+An attacker who has Kubernetes API access, has sufficient privileges to delete
+the webhook  object in the cluster.
+
+**Mitigation**
+RBAC rights are strictly controlled.
+
+**To do**
+I think most of the RBAC is not Kubewarden responsibility. But we can help our users if we:
+- Warning them in our docs and *suggest* some minimum RBAC to be used.
+- Provide a policy which detect RBAC changes and **maybe** block them.
+
+
+## Threat 5 - Attacker gets access to valid credentials for the webhook
+An attacker gains access to valid client credentials for the admission controller webhook
+
+**Mitigation**
+Webhook fails closed.
+
+Kubewarden is failed closed. Thus, we should be fine.
+
+## Threat  6 - Attacker gains access to a cluster admin credential
+
+An attacker gains access to a cluster-admin level credential in the kubernetes cluster.
+
+**Mitigation**
+N/A
+
+I cannot see how Kubewarden can help here.
+
+## Threat 7 - Attacker sniffs traffic on the container network
+An attacker who has access to the container network is able to sniff traffic
+between the API  server and the admission controller webhook.
+
+**Mitigation**
+Webhook uses TLS encryption for all traffic
+
+Kubewarden are safe. Because the webhook connections are encrypted.
+
+
+## Threat 8 - Attacker carries out a MITM attack on the webhook
+An attacker on the container network, who has access to the NET_RAW capability
+can  attempt to use MITM tooling to intercept traffic between the API server
+and admission  controller webhook.
+
+**Mitigation**
+Webhook mTLS authentication is used.
+
+**To do**
+Kubewarden should implement mutual TLS authentication
+We can add in the recommended policies from the `kubewarden-defaults` Helm
+chart a policy to drop the `NET_RAW` capability.
+
+### Threat 9 - Attacker steals traffic from the webhook via spoofing
+An attacker is able to redirect traffic from the API server which is intended
+for the admission  controller webhook by spoofing.
+
+**Mitigation**
+Webhook mTLS authentication is used.
+
+**To do**
+Kubewarden should implement mutual TLS authentication
+
+### Threat 10 - Abusing a mutation rule to create a privileged container
+An attacker is able to cause a mutating admission controller to modify a workload,
+such that  it allows for privileged container creation
+
+**Mitigation**
+All rules are reviewed and tested.
+
+**To do**
+We may came up with some tests to cover this rules reviews
+We  should carefully review a PR changing the rules in the policies deployment
+
+## Threat 11 - Attacker deploys workloads to namespaces that are  exempt from admission control
+An attacker is able to deploy workloads to Kubernetes namespaces that are exempt
+from the  admission controller configuration.
+
+**Mitigation**
+RBAC rights are strictly controlled
+
+**To do**
+I think most of the RBAC is not Kubewarden responsability. But we can help our users if we:
+- Warning them in our docs and *suggest* some minimum RBAC to be used.
+- Provide a policy which detect RBAC changes and **maybe** block them. Is this possible?
+
+
+## Threat ID 12 - Block rule can be bypassed due to missing match (e.g.  missing initcontainers)
+An attacker created a workload manifest which uses a feature of the Kubernetes
+API which  is not covered by the admission controller
+
+**Mitigation**
+All rules are reviewed and tested.
+
+**To do**
+We may came up with some tests to cover this rules reviews
+We  should carefully review a PR changing the rules in the policies deployment
+
+## Threat ID 13 - Attacker exploits bad string matching on a blocklist to  bypass rules
+An attacker, who has rights to create workloads, bypasses a rule by exploiting
+bad string  matching.
+
+**Mitigation**
+All rules are reviewed and tested.
+
+**To do**
+We may came up with some tests to cover this rules reviews
+We  should carefully review a PR changing the rules in the policies deployment
+
+## Threat ID 14 - Attacker uses new/old features of the Kubernetes API  which have no rules
+An attacker, with rights to create workloads, uses new features of the Kubernetes
+API (for  example a changed API version) to bypass a rule.
+
+**Mitigation**
+All rules are reviewed and tested.
+
+**To do**
+We may came up with some tests to cover this rules reviews and API versions.
+We can create a configuration to reject by default requests where the API
+version not cover by the policy.  We should warning policies developers to cover
+all the supported API version in theirs tests and reject all of others.
+
+
+## Threat ID 15 - Attacker deploys privileged container to node running  Webhook controller
+An attacker, who has rights to deploy privileged containers to the cluster, creates
+a privileged  container on the cluster node where the admission controller webhook operates.
+
+**Mitigation**
+Admission controller uses restrictive policies to prevent privileged  workloads
+
+**To do**
+I do not know if Kuberwarden can help on this. Kubewarden does not have access to
+containers running in the cluster node.
+
+
+## Threat ID 16 - Attacker mounts a privileged node hostpath allowing  modification of Webhook controller configuration
+An attacker, who has rights to deploy hostPath volumes with workloads, creates a
+volume  which allows for access to the admission controller pod’s files.
+
+**Mitigation**
+Admission controller uses restrictive policies to prevent privileged  workloads
+
+**To do**
+We can add a recommended policy in the `kubewarden-default` Helm chart to prevent this.
+
+
+## Threat ID 17 - Attacker has privileged SSH access to cluster node  running admission webhook
+An attacker is able to log into cluster nodes as a privileged user via SSH.
+**Mitigation**
+N/A
+
+I don't think Kubewarden can help on this
+
+
+## Threat ID 18 - Attacker uses policies to send confidential data from  admission requests to external systems
+An attacker is able to configure a policy that listens to admission requests and
+sends  sensitive data to an external system.
+
+**Mitigation**
+Strictly control external access for webhook
+
+Kubewarden policies run in a restrictive environment. They do not have network access.
+
+
+## Threat Kubewarden ID 1 - Bootstrapping of trust for admission controller
+Assuming a trusted but fresh Kubernetes cluster, an attacker is able to compromise the Kubewarden stack before any of the policies securing it is deployed and enforcing. For example, by using unsigned and malicious images for kubewarden-controller, policy-server, or any of the Kubewarden dependencies (cert-manager) or optional dependencies (grafana, prometheus..), or by compromising the Helm charts payload.
+
+** Mitigation **
+1. Kubewarden provides a Software Bill Of Materials, which lists all images needed. This aids with Zero-Trust.
+  The Kubernetes Administrator must verify the Kubewarden images (and its dependencies' images and charts) out of the Kubernetes cluster, in a trusted environment. This can be done with `cosign`, for example.
+  Incidentally, this is part of the implementation needed for air-gapped installations.
+ 2. Use signed Helm charts, and verified digests (instead of tags) for Kubewarden images in those Helm charts. This doesn't secure dependencies though.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+
+# Alternatives
+[alternatives]: #alternatives
+
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+


rfc/0006-threat-model.md
```diff
@@ -0,0 +1,262 @@
...
+Kubewarden default behavior  already does that.
+
+## Threat 2 - Attacker passes workloads which require complex  processing causing timeouts
+
+An attacker, who can access the admission controller at a network level, passes
+requests to  the admission controller which require complex processing, causing
+timeouts as the  admission controller uses compute power to process the workloads
+
+**Mitigation**
+Webhook fail closed and authenticate callers. We are safe, Kubewarden already does that.
```
raulcabello, line 60: just to mention that someone could change `failurePolicy` to `ignore`. But by default is `fail` so I agree that we are safe here.

rfc/0006-threat-model.md
```diff
@@ -0,0 +1,262 @@
...
+**To do**
+Kubewarden should implement mutual TLS authentication
+
+### Threat 10 - Abusing a mutation rule to create a privileged container
+An attacker is able to cause a mutating admission controller to modify a workload,
+such that  it allows for privileged container creation
+
+**Mitigation**
+All rules are reviewed and tested.
+
```
raulcabello, line 140: [pod-privileged-policy](https://github.com/kubewarden/pod-privileged-policy) can also be used here

raulcabello (approved):
LGTM, thanks @jvanz ! I just left some minor comments

rfc/0006-threat-model.md
```diff
@@ -0,0 +1,262 @@
...
+An attacker who has access to the Webhook endpoint, at the network level, could
+send large  quantities of traffic, causing an effective denial of service to the
+admission controller.
+
+**Mitigation**
+
+Webhook fails closed. In other words, if the webhook does not respond in time,
+for any reasion, API server should reject the request. We are safe on this.
+Kubewarden default behavior  already does that.
```
ereslibre, line 51: Given the policy-server service is reachable from within the cluster, it would be possible to attempt a denial of service by accessing the service from within the cluster, by requesting a big amount of requests.

If the policy is configured to fail open (`failurePolicy: ignore`), then policies will have no effects if the endpoint is not reachable. If the policy is configured to fail closed (`failuePolicy: fail`), then it can cause a denial of service if the webhooks are not reachable.

Network policies would mitigate this threat.

rfc/0006-threat-model.md
ereslibre: Same comment as in Threat 1.

rfc/0006-threat-model.md
```diff
@@ -0,0 +1,262 @@
...
+Kubewarden is failed closed. Thus, we should be fine.
+
+## Threat  6 - Attacker gains access to a cluster admin credential
+
+An attacker gains access to a cluster-admin level credential in the kubernetes cluster.
+
+**Mitigation**
+N/A
+
+I cannot see how Kubewarden can help here.
```
ereslibre, line 99: It is also not very clear in the original description "via a number of mechanisms.".

rfc/0006-threat-model.md
```diff
@@ -0,0 +1,262 @@
...
+can  attempt to use MITM tooling to intercept traffic between the API server
+and admission  controller webhook.
+
+**Mitigation**
+Webhook mTLS authentication is used.
+
+**To do**
+Kubewarden should implement mutual TLS authentication
+We can add in the recommended policies from the `kubewarden-defaults` Helm
+chart a policy to drop the `NET_RAW` capability.
```
ereslibre, line 122: `NET_RAW` containers will always exist though. CNI plugins are a typical example.

rfc/0006-threat-model.md
ereslibre: As @raulcabello mentions, if an explicit pod-privileged-policy is installed (validating only), then this cannot happen. Regardless of whatever the mutating policy did, the validating policy has to approve the request.

This is a harder situation for example if a user deploys a policy that does mutation and validation in the same policy (e.g. to save time reducing requests made by the api server). If a policy does many things, then it's up to the policy author to make sure that this kind of situation does not happen.

rfc/0006-threat-model.md
```diff
@@ -0,0 +1,262 @@
...
+An attacker is able to deploy workloads to Kubernetes namespaces that are exempt
+from the  admission controller configuration.
+
+**Mitigation**
+RBAC rights are strictly controlled
+
+**To do**
+I think most of the RBAC is not responsability. But we can help our users if we:
+- Warning them in our docs and *suggest* some minimum RBAC to be used.
+- Provide a policy which detect RBAC changes and **maybe** block them. Is this possible?
```
ereslibre, line 155: Also, I think is a good practice to deploy `ClusterAdmissionPolicy` policies with a reject list by default. This ensures that these policies are enforced everywhere except for a well known list of namespaces where they are not enforced.

This makes Threat 11 harder to happen.

rfc/0006-threat-model.md
```diff
@@ -0,0 +1,262 @@
...
+API (for  example a changed API version) to bypass a rule.
+
+**Mitigation**
+All rules are reviewed and tested.
+
+**To do**
+We may came up with some tests to cover this rules reviews and API versions.
+We can create a configuration to reject by default requests where the API
+version not cover by the policy.  We should warning policies developers to cover
+all the supported API version in theirs tests and reject all of others.
```
ereslibre, line 191: Yeah, this is tricky. Imagine a policy already built some time ago. Now the API server adds support for ephemeral containers, but this policy only checks init and regular containers: there's no way in which we could have warned the policy author when they wrote the policy some time ago: the Kubernetes API didn't have ephemeral containers.

I think the best we can do is document in our blog when this kind of API addition requires to review existing policies in common API's.

rfc/0006-threat-model.md
```diff
@@ -0,0 +1,262 @@
...
+
+**Mitigation**
+Admission controller uses restrictive policies to prevent privileged  workloads
+
+**To do**
+We have a policy to prevent privileged workloads in the recommended policies in
+the `kubewarden-default` helm chart. But we can improve by:
+- We may need to define all the API versions cover by the recommended policies
+- We can create a configuration to reject by default requests where the API version not cover by the policy.
+- We should warning policies developers to cover all the supported API version in theirs tests and reject all of others.
```
ereslibre: These don't apply here. Leftover?

ereslibre: We provide policies that can help with this and that can be deployed in the default `policy-server` namespace, where all policy server deployments are created.

rfc/0006-threat-model.md
```diff
@@ -0,0 +1,262 @@
...
+
+## Threat ID 18 - Attacker uses policies to send confidential data from  admission requests to external systems
+An attacker is able to configure a policy that listens to admission requests and
+sends  sensitive data to an external system.
+
+**Mitigation**
+Strictly control external access for webhook
+
+**To do**
+We may define some network configuration to block external access from the Kubewarden pods.
```
ereslibre: Policies run in a very restrictive environment, I think while we should have this threat in consideration it does not apply that much to Kubewarden.

ereslibre (commented):
Thank you @jvanz!

rfc/0006-threat-model.md
```diff
@@ -0,0 +1,262 @@
...
+# Summary
+[summary]: #summary
+
+Describe the threat model for the admissions control, how it can effect Kubewarden
+and the possible mitigations.
+
+# Motivation
+[motivation]: #motivation
+
+The goal of this is turn Kubewarden mode secure to operate.
```
viccuad:
```suggestion
The goal of this is turn Kubewarden more secure to operate.
```
(I think)

rfc/0006-threat-model.md
```diff
@@ -0,0 +1,262 @@
...
+The threat model of the Kubernetes admission control list 18 possible threats.
+All of them will be discussed sharing a brief description of it, possible mitigation
+and what we, as Kubewarden team, should do about it (if necessary). As well as
+help our users to be aware of the risks.
+
+If more details about each threat is needed, go to the original document here:
+https://github.com/kubernetes/sig-security/tree/main/sig-security-docs/papers/admission-control
+
+
+## Threat 1 - Attacker floods webhook with traffic preventing its
```
viccuad:
```suggestion
## Threat 1 - Attacker floods webhook with traffic preventing its operation
```

rfc/0006-threat-model.md
viccuad: I suppose then drop `NET_RAW` capability besides in a rejectlist ns (control plane).

rfc/0006-threat-model.md
viccuad: Still, it sounds like a gotcha that deserves a paragraph in the policy authors docs.

rfc/0006-threat-model.md
viccuad: Agree here.
I wonder if we could provide that rejectlist in a configmap deployed via `kubewarden-defaults`, that all ClusterAdmissionPolicy can use. We could even configure the default policy-server to apply that rejectlist for all ClusterAdmissionPolicies that users create, leaving the control plane and Kubewarden undisturbed.

rfc/0006-threat-model.md
```diff
@@ -0,0 +1,262 @@
...
+- Warning them in our docs and *suggest* some minimum RBAC to be used.
+- Provide a policy which detect RBAC changes and **maybe** block them. Is this possible?
+
+
+## Threat ID 12 - Block rule can be bypassed due to missing match (e.g.  missing initcontainers)
+An attacker created a workload manifest which uses a feature of the Kubernetes
+API which  is not covered by the admission controller
+
+**Mitigation**
+All rules are reviewed and tested.
```
viccuad, line 163: I expect this to be fully covered by `kubewarden-defaults` policies.

rfc/0006-threat-model.md
viccuad: Maybe we can add metadata to policies, so authors are explicitly stating which bracket of API version they cover against. We as authors can review and bump as needed, but yep, it's a cat and mouse game.

I suppose we either cover all of K8s API for all supported versions with policies. Or we create a policy that rejects all k8s objects, besides a very specific API allowlist.

rfc/0006-threat-model.md
viccuad: I suppose then drop NET_RAW capability besides in a rejectlist ns (control plane).

viccuad (approved):
LGTM, thanks :). Left some comments on the ideas, which I suppose would go into default policies, yet for me the RFC is ok.

rfc/0006-threat-model.md
flavio: I think the only reasonable way to mitigate this issue is to introduce allow each policy to run for at most X seconds. Once the timeout is reached the evaluation response will be a rejection accompanied by a certain error code and message.

rfc/0006-threat-model.md
flavio: Also in this case, I think the only reasonable way to mitigate this issue is to introduce allow each policy to run for at most X seconds. Once the timeout is reached the evaluation response will be a rejection accompanied by a certain error code and message.

rfc/0006-threat-model.md
```diff
@@ -0,0 +1,262 @@
...
+An attacker, who has rights to create workloads in the cluster, is able to exploit
+a mis-configuration to bypass the intended security control
+
+**Mitigation**
+Regular reviews of webhook configuration catch issues.
+
+**To do**
+We may provide some mechanism to notify operators when a webhook is changed.
+Maybe the controller can watch its webhooks, fix when the configuration is
+changed and notify the operator.
```
flavio: We already have a quite a list of things to do, I would not give the impression we're going to do something, unless there's a clear responsibility on our side/anything concrete we can do.

In this case, I think there's nothing we can do. The mitigation proposed by the paper makes sense. Our users can also leverage `kwctl run` to keep their configuration tested and look out for these edge cases.

rfc/0006-threat-model.md
```diff
@@ -0,0 +1,262 @@
...
+An attacker who has Kubernetes API access, has sufficient privileges to delete
+the webhook  object in the cluster.
+
+**Mitigation**
+RBAC rights are strictly controlled.
+
+**To do**
+I think most of the RBAC is not Kubewarden responsibility. But we can help our users if we:
+- Warning them in our docs and *suggest* some minimum RBAC to be used.
+- Provide a policy which detect RBAC changes and **maybe** block them.
```
flavio, line 81: I think there are two things to take into account for this scenario.

## The attacker changes one of our CR

An attacker could delete a `ClusterAdmissionPolicy` or an `AdmissionPolicy` object. There's nothing we can do in that case. We might want to highlight how important is to give the RBAC privileges to change these objects only to the users that are trusted.

## The attacker changes a WebhookConfiguration object

Other than stating how important is to control via RBAC who can "touch" these Kubernetes objects... there's one question I have: is our controller able to identify the drift between what is described inside of a `ClusterAdmissionPolicy`/`AdmissionPolicy` and its related webhook configuration object? If our controller is able to notice that, we would be able to restore the policy.
This is ofc not enough, but would be helpful.

rfc/0006-threat-model.md
flavio: We can prevent the creation of containers that have this capability via [this policy](https://github.com/kubewarden/capabilities-psp-policy). Is that one of the policies we enable via our "PSP best practices" selection?

rfc/0006-threat-model.md
flavio: If our "PSP best practices" selection deploys the pod-privileged-policy, this is not going to be an issue. Even if the user provides a mutating policy that by design/mistake makes a container privileged.

rfc/0006-threat-model.md
flavio: I agree with what rafa wrote

rfc/0006-threat-model.md
```diff
@@ -0,0 +1,262 @@
...
+
+## Threat ID 16 - Attacker mounts a privileged node hostpath allowing  modification of Webhook controller configuration
+An attacker, who has rights to deploy hostPath volumes with workloads, creates a
+volume  which allows for access to the admission controller pod’s files.
+
+**Mitigation**
+Admission controller uses restrictive policies to prevent privileged  workloads
+
+**To do**
+We can add a recommended policy in the `kubewarden-default` Helm chart to prevent this.
```
flavio, line 214: Disabling the usage of `hostPath` was one of the recommended PSP settings. Do we enforce that via our "best of PSP" selection?

rfc/0006-threat-model.md
flavio: Agree, this doesn't apply to Kubewarden policies because we don't offer network access to our Wasm policies.

rfc/0006-threat-model.md
viccuad: This comment from Flavio makes me want to have all entries in this document with "test configuration" be "test configuration with kwctl", just to be explicit.

rfc/0006-threat-model.md
```diff
@@ -0,0 +1,262 @@
...
+An attacker is able to configure a policy that listens to admission requests and
+sends  sensitive data to an external system.
+
+**Mitigation**
+Strictly control external access for webhook
+
+**To do**
+We may define some network configuration to block external access from the Kubewarden pods.
+
+
```
viccuad, line 243: I would like to add another threat. Maybe is overkill, but I think it's worth the note:

```suggestion
## Threat Kubewarden ID 1 - Bootstrapping of trust for admission controller
Assuming a trusted but fresh Kubernetes cluster, an attacker is able to compromise the Kubewarden stack before any of the policies securing it is deployed and enforcing. For example, by using unsigned and malicious images for kubewarden-controller, policy-server, or any of the Kubewarden dependencies (cert-manager) or optional dependencies (grafana, prometheus..), or by compromising the Helm charts payload.

** Mitigation **
1. Kubewarden provides a Software Bill Of Materials, which lists all images needed. This aids with Zero-Trust.
  The Kubernetes Administrator must verify the Kubewarden images (and its dependencies' images and charts) out of the Kubernetes cluster, in a trusted environment. This can be done with `cosign`, for example. 
  Incidentally, this is part of the implementation needed for air-gapped installations.
 2. Use signed Helm charts, and verified digests (instead of tags) for Kubewarden images in those Helm charts. This doesn't secure dependencies though.
 
```

rfc/0006-threat-model.md
jvanz: I think we should keep failing closed. This is a good way to avoid bad actors from deploying malicious apps during a DOS.

rfc/0006-threat-model.md
jvanz: Yes, users with permissions can change that. This is also mapped in threat #4. But I still believe we  should keep failing closed. This is a good practice to avoid bad actors from deploying malicious apps during a DOS. Even if we add some timeout, given enough load, that may not be sufficient. Furthermore, as far as I can remember, users can change this behavior with helm values.

rfc/0006-threat-model.md
jvanz: Yes, that's what I had in mind. Adds the [capabilities-psp-policy](https://github.com/kubewarden/capabilities-psp-policy) in the recommended policies skipping the control plane namespaces already listed in the chart.

rfc/0006-threat-model.md
jvanz: Yes, the "PSP best practices" includes the `pod-privileged-policy`.

rfc/0006-threat-model.md
jvanz: I think it's nice to allow our user to choose if they want to reject requests from unexpected API version or just accept the requests. As @viccuad said, this can be done using policies. Which is nice, because we do not need to change the core of the project and still give an option for our users to protect themselves from this. 

I need to check, but if the context aware policies have read access to the Kuberwarden objects, they can read the mapped API version dynamically and rejecting requests of API version that are not listed there.

rfc/0006-threat-model.md
jvanz: ~Yeah, we are already doing that in our recommended policies: https://github.com/kubewarden/helm-charts/blob/main/charts/kubewarden-defaults/templates/host-namespace-policy.yaml~

rfc/0006-threat-model.md
jvanz: :+1:

rfc/0006-threat-model.md
jvanz: Yes, copy and past issue. :man_facepalming:

jvanz (commented):
@kubewarden/kubewarden-developers , JFYI: most of the mitigations came from the official k8s document.

rfc/0006-threat-model.md
jvanz: Furthermore, our policies cannot help on this. The threat said container deployed in the cluster node. Kubewarden does not have access to this information. Image a bad actors with access to the node running the controller, we cannot block them from running a privileged container in the node.

rfc/0006-threat-model.md
jvanz: Actually the `To do` section does not seem to make sense in the context of this threat. The controller cannot know which is the "right" configuration. Thus, I cannot notify changes in the cluster that it's not taking care. This seems similar to the mitigation of other threats, so it can be a confusion during the document edit.

rfc/0006-threat-model.md
jvanz: I think the controller can compare the yaml configuration of the current webhook and the yaml which it generates from the policy definition. If it's different, it can reapply it. Cannot it?

rfc/0006-threat-model.md
jvanz: We do not install the `hostpaths-psp-policy` in the recommended policies in the `kubewarden-default` chart. So, we need to add it.

jvanz:
@kubewarden/kubewarden-developers, I've reviewed the conversations in this PR and I come up with some actionable items from all the team. Thus, we can discuss the items, change or even drop some of them. They are:


1. https://github.com/kubewarden/docs/issues/115
2. https://github.com/kubewarden/docs/issues/116
3. https://github.com/kubewarden/docs/issues/117
4. https://github.com/kubewarden/helm-charts/issues/90
5. https://github.com/kubewarden/helm-charts/issues/91

6. https://github.com/kubewarden/policy-server/issues/254
7. https://github.com/kubewarden/kubewarden-controller/issues/224
8. https://github.com/kubewarden/kubewarden-controller/issues/225
9. https://github.com/kubewarden/kubewarden-controller/issues/226
10. https://github.com/kubewarden/kubewarden-controller/issues/227
11. https://github.com/kubewarden/kubewarden-controller/issues/228
12. https://github.com/kubewarden/helm-charts/issues/92


I believe we should do items 1 to 5 now. They are easy to do.  All the other should be reviewed with more care. 

Please, let me know if I forgot something.

viccuad:
That list of action items sounds fine to me. We could create cards already for each of them and discuss them in the planning.

The PR is also complete from my point of view. Nice job!

viccuad (approved):

raulcabello (approved):
it looks great! Thanks @jvanz

ereslibre (approved):
Thanks for this @jvanz.

Although I have some doubts on how to approach some specific issues, I think the split looks great and actionable.

PR #3, opened 2022-04-22
feat: Add state property to RFCs, give state to existing

viccuad:
## Description

Add state property to RFCs, give state to existing RFCs.

Shamelessly lifted from https://dep-team.pages.debian.net/deps/dep0/.

### DIFF
--- a/rfc/0000-template.md
+++ b/rfc/0000-template.md
@@ -60,22 +60,17 @@ disruptive technical decisions (when applicable), as well as corner-cases and
 warnings.
 --->
 
-Take this template and propose an RFC by opening a PR against this repo.
+Adapt this template into a new RFC and open a PR against this repo.
 
-During the RFC life, update its state with one from:
-- `DRAFT`: Normal discussion and changes happen in this state. Achieving
-  consensus on what to be done and how. Silence is not agreement but unanimity
-  is not needed either.
-- `CANDIDATE`: Meant to prove that an implementation is feasible. Changes to the
-  implementation can happen in this period, normally from feedback from the
-  implementation.
-- `ACCEPTED`: Consensus exists that implementation has been a success.
-- `REJECTED`: No consensus either for the draft text or the implementation.
-- `OBSOLETE`: No longer relevant (overridden by other RFCs, object of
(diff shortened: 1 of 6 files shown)


viccuad:
Maybe this is too convoluted for our use case. If so, happy dropping it. Upsides for me is that it brings clarity and a foundation for community.
Also, with draft and candidate states, means that one doesn't need to finish an RFC in the first PR, but we can iterate.

rfc/0000-template.md
```diff
@@ -5,6 +5,24 @@
...
+  implementation can happen in this period, normally from feedback from the
+  implementation.
+- `ACCEPTED`: Consensus exists that implementation has been a success.
+- `REJECTED`: No consensus either for the draft text or the implementation.
+- `OBSOLETE`: No longer relevant (overridden by other RFCs, object of
+  enhancement no longer in use).
+
+Ideal progression:
+`DRAFT` (via a PR, until that PR is merged) -> `CANDIDATE` (iterate implementation) -> `ACCEPTED` (implementation is a sucess).
+
```
jvanz: Is not the  metadata table the best place to the status field? I think it makes sense to keep it together with the category, pr, etc...

viccuad: yep, you are right! Changed it. Also, changed the template, so the state explanation is part of its description.

jvanz (approved):
LGTM! :D

ereslibre:
I think https://dep-team.pages.debian.net/deps/dep0/ is too complex to adapt as a complete workflow. I would keep `Accepted` and `Rejected` state only.

By default, when a RFC is created and a PR is opened, it is created with `Accepted` as the status. After discussion and adapting, if it merges, it is merged as `Accepted`.

If it is finally rejected or consensus is not reached, it is changed to `Rejected` and merged.

I would not go much further than that to be honest. Having a record of the RFC's that have been rejected in the tree is good enough for us I think.

viccuad:
Yes, indeed, I went overboard. Simplified verbatim with your approach Rafa!

ereslibre (approved):
LGTM, thanks @viccuad

PR #4, opened 2022-05-05
Single version proposal

### LINKED ISSUE #5: RFC: have a unified version for the whole Kubewarden stack
We would like to have a way to easily identify the version of Kubewarden a user is installing/running. Right now each component of kubewarden has its own different version number.

## Acceptance criteria

* Create a RFC that proposes a way to version Kubewarden that simplifies life of users, supporters and marketing

flavio:
A new take on this problem, as discussed with the rest of the team.

### DIFF
--- a/rfc/0006-kubewarden-versioning.md
+++ b/rfc/0006-kubewarden-versioning.md
@@ -1,0 +1,210 @@
+|              |                                                 |
+|:-------------|:------------------------------------------------|
+| Feature Name | Kubewarden versioning                           |
+| Start Date   | Mar 3rd 2022                                    |
+| Category     | Versioning                                      |
+| RFC PR       | [PR4](https://github.com/kubewarden/rfc/pull/4) |
+
+# Summary
+[summary]: #summary
+
+Kubewarden is composed of many core components that work together in
+order to build the final product. This is a proposal to harmonize
+Kubewarden versioning schema.
+
+# Motivation
+[motivation]: #motivation
+
+Having a way to specify what version of Kubewarden is being installed
+or executed would make it easier for the Kubewarden team to describe,
+market and support Kubewarden (as in user support).
+
+It would help users in understanding what version of Kubewarden they
+are about to install or execute.
+
+## Examples / User Stories
+[examples]: #examples
+
+### User story #1
+
+> Users do not need to keep track of component versions and how they
+> work together. They just need to care about running a specific
+> Kubewarden version.
+
+### User story #2
+
+> Users are able to easily understand that a new version of Kubewarden
+> is available, and that more than one component might be upgraded in
+> order to move to the new Kubewarden version.
+
+### User story #3
+
+> A user has an issue with Kubewarden. Given its multiple component
+> nature, we would like to know what was the Kubewarden version that
+> got installed, and through what versions it was upgraded.
+
+### User story #4
+
+> As Kubewarden maintainers we want to define a version skew policy
+> that is easy to be understood and implement.
+> Something similar to what [Kubernetes does](https://kubernetes.io/releases/version-skew-policy/)
+
+# Detailed design
+[design]: #detailed-design
+
+## Many components, one experience
+
+Kubewarden core components, which are exposed to the user are:
+
+- Helm Charts
+- Kubewarden Controller
+- Policy Server
+- kwctl
+
+As a second dimension, there are also policies and their respective
+SDK's. This proposal is going to ignore this dimension on purpose,
+because Kubewarden runtime components (`policy-server` and `kwctl`)
+should be compatible with policies previously built, even with older
+language SDK's.
+
+We want to have a single version for these "public" Kubewarden
+components. This version is going to be expressed using
+[Semantic Versioning](https://semver.org/).
+
+The Kubewarden Controller, the Policy Server and kwctl components are going
+to share the same *Major* and *Minor* version numbers. Each project is
+going to have an independent *Patch* level number.
+
+All the non-visible components of Kubewarden (e.g. the different
+Rust libraries) and the policies are going to retain their
+existing version numbers, release strategy and, generally speaking,
+independence.
+
+### Helm charts
+
+Helm charts have two kind of version numbers:
+
+* `version`: a SemVer 2 version specific to the helm chart
+* `appVersion`: the version of the app that the chart contains
+
+The helm charts will keep their own independence when it comes to
+the `version` attribute.
+On the other hand, the `appVersion` attribute
+will always be set to the `<Major>.<Minor>` version of the
+Kubewarden stack.
+
+> See the official documentation about
+[`Chart.yaml`](https://helm.sh/docs/topics/charts/#the-chartyaml-file)
+for more information.
+
+## Examples
+
+This section provides clarity about how the proposal would work
+using some concrete real-world scenarios.
+
+### A new release Kubewarden takes place
+
+A new version of Kubewarden stack has to be released. That can
+happen because a new important feature/bugfix has been added,
+because of marketing reasons,...
+
+Let's assume:
+
+* The current version of the Kubewarden stack is `0.5.3`
+* The current version of the helm chart is `v0.10.3`
+
+The new feature is not worth a major bump of the "unique kubewarden stack version",
+hence we will perform a minor update.
+
+We are going to tag `v0.6.0` of all these components and release them:
+
+* Kubewarden Controller
+* Policy Server
+* kwctl
+
+This will happen **regardless** of the changes done to these projects. For
+example, kwctl is going to have a new release `v0.6.0` even if it didn't
+have any commits since `v0.5.3` was tagged.
+
+The Helm charts are updated in this way:
+
+* The version of container images (Policy Server, Kubewarden
+  Controller) are set to be `0.6.0`
+* The chart `version` attribute is set to `v0.10.4` because no significant
+  changed happened
+* The `appVersion` attribute is set to `0.6.0`
+
+### A patch is done to kwctl
+
+A series of minor issues are found inside of kwctl. We decide to
+tag a new release that includes these fixes.
+
+The version of `kwctl` is updated from `0.6.0` to `0.6.1`.
+No other releases are done
+
+### A patch is done to Kubewarden Controller
+
+A minor issue is fixed with Kubewarden Controller. We decide to
+tag a new release that includes the fix.
+
+The following actions are done:
+
+* We tag a new release of kubewarden-controller: `0.6.1`
+* We update the helm chart that makes use of this image:
+  * We update the references of the image to ensure `0.6.1` is being pulled
+  * We update the `version` number of the chart itself to be
+    change from `0.10.4` to `0.10.5`.
+    
+    **Note:** we don't change the `appVersion` attribute
+
+### A major security issue is found inside of Policy Server
+
+A major security issue is found inside of Policy server.
+Changes are done to both policy server and Kubewarden Controller.
+
+We decide to release a new version of the whole stack, because of
+the size of the changes and the impact. We will bump the
+Kubewarden stack from `0.6.0` to `0.7.0`.
+
+We proceed with the following actions:
+
+* Tag a new release of Kubewarden Controller: `0.7.0`
+* Tag a new release of Policy Server: `0.7.0`
+* Tag a new release of kwctl: `0.7.0`. We do that even though
+  no actual change has been done to kwctl
+* helm chart:
+  * Update all the references to the Kubewarden Controller and
+    Policy Server images
+  * Update `version`: this time the changes involve some updates to
+    the helm chart itself. We deem this is worth a minor update.
+    Because of that, the `version` goes from `0.10.5` to `0.11.0`
+  * Update the `appVersion` to be `0.7.0`
+
+### A bug is found inside of one of the helm charts
+
+A small fix is done to one of the existing helm charts.
+
+The version of the Kubewarden stack is **not** going to be changed.
+Only the `version` inside of the patched helm chart is going to be
+bumped.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- There are times when the version of one of the components of
+  the "public stack" is bumped even if no actual changes have
+  been done to its source code.
+- The global stack version does not make use of the `Patch`
+  attribute of Semantic Versioning. That's because we want to
+  leave some freedom to each component of the "public stack".
+
+# Alternatives
+[alternatives]: #alternatives
+
+[This](https://github.com/kubewarden/kubewarden-controller/pull/182)
+rejected proposal.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+None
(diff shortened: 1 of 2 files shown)


rfc/0006-kubewarden-versioning.md
```diff
@@ -0,0 +1,226 @@
...
+The following actions are done:
+
+* We tag a new release of kubewarden-controller: `0.6.1`
+* We update the helm chart that makes use of this image:
+  * We update the references of the image to ensure `0.6.1` is being pulled
+  * We update the `version` number of the chart itself to be
+    change from `0.10.4` to `0.10.5`.
+    
+    **Note:** we don't change the `appVersion` attribute
+
```
jvanz: Okay, if I understand correctly we could end with multiple users using the same `appVersion` (e.g. `0.6.0`) but actually running different container images. Right?

rfc/0006-kubewarden-versioning.md
```diff
@@ -0,0 +1,226 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+[This](https://github.com/kubewarden/kubewarden-controller/pull/182)
+rejected proposal.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
```
jvanz: How to execute the release process? As we have a bunch o different versions the release process can be very error prone. Therefore, I believe we should try to automate or, at least, document **very** well how to do it.

rfc/0006-kubewarden-versioning.md
raulcabello: yes with different patch level images. We need this if we want each project to have an independent patch level number

rfc/0006-kubewarden-versioning.md
```diff
@@ -0,0 +1,226 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- There are times when the version of one of the components of
+  the "public stack" is bumped even if no actual changes have
+  been done to its source code.
+- The global stack version does not make use of the `.patch`
+attribute of Semantic Versioning. That's because we want to
+leave some freedom to each component of the "public stack".
```
raulcabello: I would add that we could end with multiple users using the same appVersion (e.g. 0.6.0) but actually running different container images as @jvanz mentioned

rfc/0006-kubewarden-versioning.md
raulcabello: I would prefer to automate the whole process

rfc/0006-kubewarden-versioning.md
```diff
@@ -0,0 +1,226 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+[This](https://github.com/kubewarden/kubewarden-controller/pull/182)
+rejected proposal.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+None
```
raulcabello: I assume this versioning is for the `kubewarden-controller` helm chart, right? `kubewarden-crds` and `defaults` will have their own versions?

rfc/0006-kubewarden-versioning.md
kkaempf: How would users then report the version (of each component) in a bug report ? Reporting the chart version or the appVersion is not sufficient.
Note that users might only know the GUI, without ever having seen `kwctl`.

flavio: If the user tells us the `version` of the helm chart he has installed, we will be able to determine the exact versions of the container images being deployed.

rfc/0006-kubewarden-versioning.md
flavio: We already have automation at the single project level, we would need something that orchestrates that.

I would however wait a little before investing time into doing that to understand:
  * If this versioning proposal sound, are we going to face unexpected issues?
  * How often does that release happen? What is the tradeoff between investing time into automating the "high level orchestrator" vs following a clearly written playbook?

rfc/0006-kubewarden-versioning.md
flavio: Something went wrong with my rebase and GitHub comments (I even see some of them repeated twice)...
I this comment lost track of the original text, I'll try to infer what you meant...

Each helm chart has its own `version`, they all the the same `appVersion`

flavio:
Another possible solution would be to make use of the `patch` attribute of the universal stack version.

Pros:
  * All the components have the same version. The `appVersion` inside of the helm chart is reflecting the version of all the container images

Cons:
  * A small fix to a single component triggers the release of all the others.
  * Users might get the feeling to update their cluster because a new patch release has been tagged, but actually no change has been done to the components running inside of the cluster.

We would need to invest more time and efforts into automating the whole release "overlord orchestrator" or just build up fixes and push them once they reach a certain mass.

Honestly, I have mixed feelings about this approach. That why I excluded that from the proposal. **However**, we could also start with this proposal and later make use of the `patch` attribute.

rfc/0006-kubewarden-versioning.md
```diff
@@ -0,0 +1,210 @@
+|              |                                                 |
```
viccuad: Nit: I would number it 0007, and have the [rejected one](https://github.com/kubewarden/kubewarden-controller/pull/182) present in the repo as 0006.

rfc/0006-kubewarden-versioning.md
```diff
@@ -0,0 +1,210 @@
...
+
+The following actions are done:
+
+* We tag a new release of kubewarden-controller: `0.6.1`
+* We update the helm chart that makes use of this image:
+  * We update the references of the image to ensure `0.6.1` is being pulled
+  * We update the `version` number of the chart itself to be
+    change from `0.10.4` to `0.10.5`.
+    
+    **Note:** we don't change the `appVersion` attribute
```
ereslibre: Why? What is the advantage of not doing so? I think having the `appVersion` of the helm chart define the version of the controller being deployed is good for traceability. Specially when we had to change the Helm chart in any way in this case.

rfc/0006-kubewarden-versioning.md
```diff
@@ -0,0 +1,210 @@
...
+
+### A major security issue is found inside of Policy Server
```
ereslibre: An interesting second take of this one is:

**A major security issue is found inside of Policy Server**: if the controller does not need any adaptation, only the Policy Server and the default Helm chart have to be updated.

An open question in this case is how we are going to bump the existing policy-server images: this is a problem regardless of the single version/umbrella versions proposal though.

rfc/0006-kubewarden-versioning.md
```diff
@@ -0,0 +1,210 @@
...
+> A user has an issue with Kubewarden. Given its multiple component
+> nature, we would like to know what was the Kubewarden version that
+> got installed, and through what versions it was upgraded.
+
+### User story #4
+
+> As Kubewarden maintainers we want to define a version skew policy
+> that is easy to be understood and implement.
+> Something similar to what [Kubernetes does](https://kubernetes.io/releases/version-skew-policy/)
+
```
viccuad: I would add:

```suggestion
### User story #5

> As a user upgrading Kubewarden to a new version, I want to know if the new version introduces backward incompatible changes and behavior, so I can decide if to upgrade, and how. For example, if upgrading the helm charts in my cluster has backwards-incompatible changes, as I may be forced to redeploy from scratch, halt workloads of do manual tasks. Or if kwctl introduces backwards-incompatible changes, that would necessitate changes to my CI infrastructure.
```

rfc/0006-kubewarden-versioning.md
ereslibre: I think any release of the controller (even patch level) forcing us to release a new helm chart version (patch level) is a good exception. Because the most important goal of the reason of the helm chart existence is to deploy the controller, along with other things, but mainly the controller.

ereslibre (commented):
Thanks, some comments.

rfc/0006-kubewarden-versioning.md
```diff
@@ -0,0 +1,210 @@
...
+[design]: #detailed-design
+
+## Many components, one experience
+
+Kubewarden core components, which are exposed to the user are:
+
+- Helm Charts
+- Kubewarden Controller
+- Policy Server
+- kwctl
```
viccuad:
```suggestion
- kwctl
- CRDs
```

(they are part of the Controller, but tracked by their own version?)

rfc/0006-kubewarden-versioning.md
```diff
@@ -0,0 +1,210 @@
...
+# Detailed design
+[design]: #detailed-design
+
+## Many components, one experience
+
+Kubewarden core components, which are exposed to the user are:
+
+- Helm Charts
+- Kubewarden Controller
+- Policy Server
```
viccuad: I'm not sure if we mean the policy-server, or the CRDs. I suppose the former, as AdmissionPolicies are not named. If so, I would name it policy-server.

rfc/0006-kubewarden-versioning.md
```diff
@@ -0,0 +1,210 @@
...
+
+### Helm charts
+
+Helm charts have two kind of version numbers:
+
+* `version`: a SemVer 2 version specific to the helm chart
+* `appVersion`: the version of the app that the chart contains
+
+The helm charts will keep their own independence when it comes to
+the `version` attribute.
```
viccuad: Nit: maybe worth pointing. Maybe it is outside of this RFC anyways.

```suggestion
the `version` attribute. That is, using semver, which for helm charts helps users be aware of backwards-incompatible changes when upgrading.
```

rfc/0006-kubewarden-versioning.md
```diff
@@ -0,0 +1,210 @@
...
+## Examples
+
+This section provides clarity about how the proposal would work
+using some concrete real-world scenarios.
+
+### A new release Kubewarden takes place
+
+A new version of Kubewarden stack has to be released. That can
+happen because a new important feature/bugfix has been added,
+because of marketing reasons,...
```
viccuad: (opening thread here, loosely related)

I get the marketing reasons, but then, it's not semver I suppose. As a user, the most valuable info for me is if there are backwards-incompatible changes. I want to know if I can blindly upgrade Kubewarden, without needing to redeploy, halt cluster workloads, or make changes.

Normally that would be provided by the helm chart, but the helm chart would get its `version` bumped a major version regardless.

rfc/0006-kubewarden-versioning.md
```diff
@@ -0,0 +1,210 @@
...
+  * Update the `appVersion` to be `0.7.0`
+
+### A bug is found inside of one of the helm charts
+
+A small fix is done to one of the existing helm charts.
+
+The version of the Kubewarden stack is **not** going to be changed.
+Only the `version` inside of the patched helm chart is going to be
+bumped.
+
```
viccuad: I would add an example about a new version that introduces a backwards-incompatible change, for kwctl, or for the charts, and see how it plays with the versioning.

rfc/0006-kubewarden-versioning.md
```diff
@@ -0,0 +1,210 @@
...
+happen because a new important feature/bugfix has been added,
+because of marketing reasons,...
+
+Let's assume:
+
+* The current version of the Kubewarden stack is `0.5.3`
+* The current version of the helm chart is `v0.10.3`
+
+The new feature is not worth a major bump of the "unique kubewarden stack version",
+hence we will perform a minor update.
```
jvanz: What do we consider a change that worth a major bump of the "unique Kubewarden stack version" ?

rfc/0006-kubewarden-versioning.md
flavio: > I get the marketing reasons, but then, it's not semver I suppose. As a user, the most valuable info for me is if there are backwards-incompatible changes. I want to know if I can blindly upgrade Kubewarden, without needing to redeploy, halt cluster workloads, or make changes.

I've been in this kind of conversations in the past (the major number because of marketing)... Most of the times there's nothing we can do about that. However, major version can happen also when a BIG change (that doesn't break the API) is introduced.

rfc/0006-kubewarden-versioning.md
flavio: I think it depends a lot on the cases... But I wound say that something like sigstore integration could have been something worth a major bump (at least for marking reasons I suspect)

rfc/0006-kubewarden-versioning.md
jvanz: The impression that I get is, in the end of the day, the only version that it will be used  is the helm chart version. Because this is the version important to solve the issues. The `appVersion` will be used only for marketing and announcements.

rfc/0006-kubewarden-versioning.md
flavio: I kept track of this conversation inside of the RFC

rfc/0006-kubewarden-versioning.md
flavio: added

rfc/0006-kubewarden-versioning.md
flavio: > Why? What is the advantage of not doing so? I think having the appVersion of the helm chart define the version of the controller being deployed is good for traceability. Specially when we had to change the Helm chart in any way in this case.

That happens because the proposal doesn't want to make use of the patch attribute inside of the global version number. The main advantage is to not be forced to tag and release all the components when there's a minor fix applied to one of them.

flavio: > I think any release of the controller (even patch level) forcing us to release a new helm chart version (patch level) is a good exception. Because the most important goal of the reason of the helm chart existence is to deploy the controller, along with other things, but mainly the controller.

The `appVersion` is about the whole stack. It's not about the kubewarden-controller, it's also about the policy server and the CRDs. If we don't want to make use of the patch level inside of the global version, then we have to avoid its usage also inside of the `appVersion` field of the helm chart.

rfc/0006-kubewarden-versioning.md
flavio: Added

rfc/0006-kubewarden-versioning.md
flavio: done

rfc/0006-kubewarden-versioning.md
flavio: good point, I've added them to the list, which means they will follow the global version number too

flavio: feel free to reopen the conversation if you do not agree with this proposal and you want instead to track them with an independent version

rfc/0006-kubewarden-versioning.md
flavio: this should be clarified now

rfc/0006-kubewarden-versioning.md
flavio: done

flavio:
@kubewarden/kubewarden-developers I've updated the document to reflect the feedback emerged during the review process. I've resolved all the comments you left over there.

For the ones that are still open, I prefer to have the commenter close the issue as solved or provide more feedback.

Thanks to everybody

viccuad (approved):
LGTM, modulo the conversations still open.

jvanz (approved):

rfc/0006-kubewarden-versioning.md
ereslibre: Thanks for the explanation.

ereslibre (approved):

PR #6, opened 2022-05-23
Upgrade statement RFC.

jvanz:
Adds a RFC describing how the upgrade process of the Kubewarden stack will work.

### DIFF
--- a/rfc/0007-upgrade-statement.md
+++ b/rfc/0007-upgrade-statement.md
@@ -1,0 +1,95 @@
+|              |                                  |
+| :----------- | :------------------------------- |
+| Feature Name | Upgrade Statement                |
+| Start Date   | 05/23/2022                       |
+| Category     | updates and versioning           |
+| RFC PR       | https://github.com/kubewarden/rfc/pull/6  |
+| State        | **ACCEPTED**                  |
+
+
+# Summary
+[summary]: #summary
+
+This RFC proposes how the upgrade of the Kuberwarden stack will work and how to do it.
+
+# Motivation
+[motivation]: #motivation
+
+Kubewarden is reaching the v1.0 milestone and we need to define a upgrade process
+to users and for the development team to follow. With it, users will be able
+to plan when and how to upgrade their Kuberwarden stack. And the Kuberwarden developers
+will be able to define when and how ship new features and fixes.
+
+## Examples / User Stories
+[examples]: #examples
+
+As Kubewarden user, I want to know, among all Kubewarden components, what are the
+compatible versions.
+
+As Kubewarden user, I want to know the upgrade path of the Kuberwarden components.
+
+As a Kubewarden developer, I want to know what is the upgrade path between
+Kubewarden versions, so that I can perform proper QA before releasing a new
+version of the stack.
+
+# Detailed design
+[design]: #detailed-design
+
+This document describes how Kubewarden versioning works and how users can upgrade
+their stack. The Kubewarden project uses [Semantic versioning](https://semver.org/)
+to define the version of all its components. In other words, the version follows
+the `MAJOR.MINOR.PATCH` pattern.  The supported version is only the latest release.
+
+The Kubewarden components that follow the rules described here are:
+
+- `kubewarden-controller`
+- `policy-server`
+- `kwctl`
+
+`kubewarden-controller`, `policy-server` and `kwctl` share the same `MAJOR`
+and `MINOR` version. The `PATCH` version can increase independently though.
+
+## Version compatibility among components
+
+`kubewarden-controller`, `policy-server` and `kwctl` should run the same `MAJOR`/`MINOR`.
+Therefore, if the `kubewarden-controller` version running is `1.1.x`, the
+`policy-server` and `kwctl` version in use should be `1.1.x` as well.
+
+## Upgrade paths
+
+When upgrading components, it is allowed to upgrade multiple `PATCH` version
+in a single shoot. However, the upgrade of multiple `MAJOR` or `MINOR` versions
+in a single upgrade is **not** supported.
+
+For example, the user is allowed to upgrade components from version `1.1.10` to
+`1.1.50` in a single upgrade. But the upgrade from `1.1.10` to `1.5.0` is not supported.
+In these cases, the user must upgrade individually to each `MAJOR`/`MINOR` version
+between the two versions. Therefore, it's necessary to upgrade `1.1.10` to `1.2.0`
+then `1.3.0` then `1.4.0` and finally to `1.5.0`. Users that want to upgrade one
+`MAJOR` version to another, also need to follow all the `MINOR` updates between the
+two `MAJOR` versions.
+
+## Upgrade order
+
+Kuberwarden users should upgrade the stack starting in the `kubewarden-controller`.
+After that, the `policy-server` and `kwctl` can be upgraded.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+
+# Alternatives
+[alternatives]: #alternatives
+
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+When upgrading major version, is it necessary upgrade minor version individually as well?
+
+Do we really want to keep the two latest releases under active maintenance?
+I think just keeping the latest version is enough for now.
+
+
+# References
+
+https://github.com/kubewarden/rfc/pull/2#issuecomment-1128965121


rfc/0008-upgrade-statement.md
```diff
@@ -0,0 +1,103 @@
...
+## Examples / User Stories
+[examples]: #examples
+
+As Kubewarden user, I want to know, among all Kubewarden components, what are the
+compatible versions.
+
+As Kubewarden user, I want to know the upgrade path of the Kuberwarden components.
+
+As Kubewarden developer, I want to know which components I need to upgrade in order
+to release a feature of a bug fix.
```
flavio: I think this is more about the single versioning story. As a Kubewarden developer, I think a better story could be:

> As a Kubewarden developer,
> I want to know what is the upgrade path between Kubewarden versions,
> so that I can perform proper QA before releasing a new version of the stack.

rfc/0008-upgrade-statement.md
```diff
@@ -0,0 +1,103 @@
...
+This means that if the latest version released of the Kubewarden components is
+`1.2.0`. Both `1.2.0` and `1.1.0` version are under active maintenance.
+
+The Kubewarden components that follow the rules described here are:
+
+- `kubewarden-controller`
+- `policy-server`
+- `kwctl`
+- Policies
+- SDKs
```
flavio: I think these ones should be left out. They are not going to follow the unique versioning number.

rfc/0008-upgrade-statement.md
```diff
@@ -0,0 +1,103 @@
...
+## Version compatibility among components
+
+`kubewarden-controller`, `policy-server` and `kwctl` should run the same `MAJOR`/`MINOR`.
+Therefore, if the `kubewarden-controller` version running is `1.1.x`, the
+`policy-server` and `kwctl` version in use should be `1.1.x` as well.
+
+Considering that `kwctl` and `policy-server` should run policies from older versions.
+Policies and the SDKs used to write them are independent from the other components.
+Their version does not need to be synchronized with the remain of the components.
+
```
flavio: JFYI: the API between between a policy and the host is already versioned. We are currently running `v1`. If we need to break the API (let's say, we decide to encode the payload using msgpack instead of JSON), we will create a `v2` version of the protocol and then we will decide what to do.

For a certain amount of time policy-server/kwctl will support both v1 and v2, later on we might decide to drop support for `v1`  and just refuse to start if a `v1` policy is provided by the user

rfc/0008-upgrade-statement.md
```diff
@@ -0,0 +1,103 @@
...
+[drawbacks]: #drawbacks
+
+
+# Alternatives
+[alternatives]: #alternatives
+
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+Do we need to define a release pace? E.g. every two weeks, every month, on demand.
```
flavio: I think this is out of the scope of this RFC

rfc/0008-upgrade-statement.md
```diff
@@ -0,0 +1,103 @@
...
+
+When upgrading major version, is it necessary upgrade minor version individually as well?
```
flavio: Can you elaborate on that please?

rfc/0008-upgrade-statement.md
```diff
@@ -0,0 +1,103 @@
...
+
+Do we really want to keep the two latest releases under active maintenance?
+I think just keeping the latest version is enough for now.
```
flavio: I think this is reasonable right now. What do the others think about it?

rfc/0008-upgrade-statement.md
```diff
@@ -0,0 +1,103 @@
...
+
+# Detailed design
+[design]: #detailed-design
+
+This document describes how Kubewarden versioning works and how users can upgrade
+their stack. The Kubewarden project uses [Semantic versioning](https://semver.org/)
+to define the version of all its components. In other words, the version follows
+the `MAJOR.MINOR.PATCH` pattern.  The supported versions are the two latest releases.
+This means that if the latest version released of the Kubewarden components is
+`1.2.0`. Both `1.2.0` and `1.1.0` version are under active maintenance.
```
raulcabello: I would keep just the latest release under active maintenance

rfc/0008-upgrade-statement.md
raulcabello: yes, I would keep just the latest version under active maintenance

rfc/0008-upgrade-statement.md
raulcabello: There might be policies that needs a new version of the policy server. Imagine we add a new host capability, then we create a policy that uses it. It will work just with the version of the policy server that contains this change, and it will not work with previous version. E.g verify-image-signatures policy won't work with old versions of policy-server.

I think it should be fine if we just mention in the policy which version of Kubewarden you need. What do you think?

flavio: > I think it should be fine if we just mention in the policy which version of Kubewarden you need. What do you think?

Yes, this is something we can mention inside of the policy. Maybe, but this is not related to this RFC, we could add a new entry inside of the policy metadata to specify which version of the host capabilities is needed. Then, at load time, policy-server/kwctl can fail if it cannot satisfy this requirement

rfc/0008-upgrade-statement.md
```diff
@@ -0,0 +1,103 @@
...
+# Summary
+[summary]: #summary
+
+This RFC proposes how the upgrade of the Kuberwarden stack will work and how to do it.
+
+# Motivation
+[motivation]: #motivation
+
+Kubewarden is reaching the v1.0 milestone and we need to define a upgrade process
+to users and for the development team to follow. Therefore, users will are be able
```
viccuad:
```suggestion
to users and for the development team to follow. With it, users will be able
```

rfc/0008-upgrade-statement.md
viccuad: me too, to be fair. I don't want to backport fixes to previous releases yet.

rfc/0008-upgrade-statement.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Version compatibility among components
+
+`kubewarden-controller`, `policy-server` and `kwctl` should run the same `MAJOR`/`MINOR`.
+Therefore, if the `kubewarden-controller` version running is `1.1.x`, the
+`policy-server` and `kwctl` version in use should be `1.1.x` as well.
+
+Considering that `kwctl` and `policy-server` should run policies from older versions.
+Policies and the SDKs used to write them are independent from the other components.
+Their version does not need to be synchronized with the remain of the components.
```
viccuad:
```suggestion
Their version does not need to be synchronized with the rest of the components.
```
(or "remainder"?)

rfc/0008-upgrade-statement.md
viccuad: Agree, just a version under active maintenance, to not need to backport security fixes.

viccuad (approved):
LGTM, many thanks! Minor nitpicks, plus only 1 version under active maintenance.

rfc/0008-upgrade-statement.md
jvanz: If the user want to upgrade from `1.0.0` to `2.0.0`. Will the upgrade be `1.0.0 -> 1.1.0 -> 1.2.0 -> any other minor version -> 2.0.0`? Or user can upgrade from `1.0.0` to `2.0.0`?

From what I wrote in the text, the user is required to follow all the minor releases. But I would like to ensure with the team that is what we want. Is it? If it is, I'll update the text to make that explicit.

rfc/0008-upgrade-statement.md
jvanz: :+1:

rfc/0008-upgrade-statement.md
jvanz: :+1:

flavio (approved):

flavio:
I think this is good to go. Once we clarify the RFC about the single versioning, I propose to publish the contents of this RFC inside of our website

rfc/0008-upgrade-statement.md
jvanz: As we removed the policies and SDKs as components from this RFC. I'll remove this paragraph.

rfc/0008-upgrade-statement.md
jvanz: I've added a sentence to make this explicit.

PR #10, opened 2022-08-09
Audit checks RFC.

### LINKED ISSUE #8: RFC: Compliance audit checks
Currently Kubewarden can be used to evaluate Kubernetes resources only when a relevant operation happens: creation, update, deletion of the resource. 

Policies evolve over time; more policies can be added or they can be redeployed with different settings. As a result of that, a Kubernetes resource that was marked as valid even a couple of hours ago, could be instead rejected later on because of a policy change. 

We want to implement a background check that inspects all the resources that are already present inside of the cluster and flags all the ones that are violating at least one of the currently enforced policies. 

The cluster operators can then instruct alerts (maybe via a Prometheus integration) to be notified when an existing resource is flagged in such a way.

We can take a look at what [Kyverno](https://kyverno.io/docs/writing-policies/background/) and [Gatekeeper](https://open-policy-agent.github.io/gatekeeper/website/docs/audit) are currently doing to solve this problem.

jvanz:
Adds the RFC to discuss how to implement the audit checks.


jvanz:
After creating this draft PR I've realized that I missed to add a text about how the audit checks will work with multiple policy servers. I'll add that before marking it as ready for review.

rfc/0009-audit-checks.md
```diff
@@ -0,0 +1,218 @@
...
+    required_drop_capabilities:
+    - NET_ADMIN
+```
+
+## Policy server and policies communication
+
+To be able to call policies to evaluate the resources already in the cluster,
+it's necessary to add a new fields in the `ValidationRequest` (e.g. `resources`)
+and `ValidationRespose` (e.g. `resources_validations`). In the `ValidationRequest`
+the field is needed to store  a list of the resources which the policy must validate.
```
raulcabello: why do we need these new fields? Can't we just scan all the resources from the policies? I mean if a policy is evaluating pods and deployments I would scan all pods and deployments for this policy. If a different policy is evaluating just configmaps I would scan all configmaps for this policy.

rfc/0009-audit-checks.md
```diff
@@ -0,0 +1,218 @@
...
+be batched and multiple instances of the same policy could be run at the same time
+temporally to speed up the policy validations.  This horizontal scaling for all the
+policies under demand is a good topic for another RFC or idea discussion for the future.
+
+
+Run the audit checks in the same policy server where the admissions requests are
+evaluated. And use cronjob to trigger audit checks. But considering [that cronjob
+can sometimes trigger none or twice a job](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#cron-job-limitations),
+plus some benefits like avoiding competing resources, a separated policy server seems a better idea
+
```
raulcabello: Another alternative similar to this one:
Run the audit policy server in a cron job. Then we don't overload existing PolicyServers. It's rare that cronjobs are triggered none or twice, but if that happens it wouldn't be such a big deal. If it not triggered it would be trigger next time, and if it is triggered twice 2 reports would be created. I don't think this will happen very often (if it happens at all)
I don't like the idea of having one policy server waiting for audit reports that might happen once a day. I think CronJobs are a good use case for generating [reports](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#cronjob)
Another good point of this approach is that a user could manually create a job and generate a report at any time easily.

raulcabello:
This is looking great @jvanz, thanks! I left some ideas

rfc/0009-audit-checks.md
jvanz: I think It's doable. But if we want to use some kind of mechanism similar to what we have for context aware policies we need to add every resource type in the allowed resources and maintain it. Furthermore, we need to add a cache in the middle. Because I would like to avoid a bad policy hammering the kubernetes API. Does it make sense?

rfc/0009-audit-checks.md
jvanz: Okay. In my first thoughts I believed that a missing audit checks could be a big deal. But if it is not, and the majority of the team agree up on this, I'm fine with using cronjobs. Actually, while I was writing this RFC this was one of my first options. But I've changed mind later

viccuad:
As talked in planning, moving to blocked to get feedback from Flavio.

rfc/0009-audit-checks.md
```diff
@@ -35,67 +35,50 @@ As a Kubernetes developer, I want access to the audit checks results to allow me
...
-```
+`PolicyServer` needs to store successful evaluation from all policies, and also failed evaluations for 
+policies in monitor mode. Failed evaluations for policies in enforce mode will not be stored as resources won't be 
+created in the cluster. These evaluations will be stored in a `PolicyReport`. There is one `PolicyReport` per namespace and 
+one `ClusterPolicyReport` for cluster-wide resources, see this [RFC](https://github.com/kubewarden/rfc/pull/13/files) for more info.
+
+Many more calls will be made to `PolicyServer` for the background checks, therefore we need to implement a cache
+for policy evaluations. The result of each policy evaluation will be stored in the cache, it will contain the `resourceVersion`
+of the policy and the resource to see if any of them have changed. Before evaluating a new resource it will check in the 
+cache if evaluation was previously done.
```
viccuad: what do we do with context-aware policies or those that make external calls?
Also, [raised in the PolicyReport RFC](https://github.com/kubewarden/rfc/pull/13#discussion_r965701469), it was said that not all requests will be audited (those performing UPDATEs for example can't be faked).

I think it would be a good idea to list specifically which kind of requests, and which kind of policies are covered by the audit scan.

rfc/0009-audit-checks.md
```diff
@@ -35,67 +35,50 @@ As a Kubernetes developer, I want access to the audit checks results to allow me
...
 Triggering another request to the Kubernetes API.
 
-It necessary a configuration to define the *time to live* period used by the cache.
-
 ## ClusterAdmissionPolicy and AdmissionPolicy CRDs
 
-It's necessary a change in the policy CRDs to mark a policy to be run in the audit checks.
-The field would be called `audit` and its default value will be `true`.
+It's necessary a change in the policy CRDs to mark a policy to be run in the background audit checks.
+The field would be called `backgroundAudit` and its default value will be `true`.
```
viccuad: The same way that policies can be `mutating` or not, and `spec.mutating` is required, and listed in the policy metadata, I think we should do the same with `spec.backgroundAudit`. This allows for correct defaults on context-aware policies (they may not be audited? dunno).
What happens if a non-auditable policy gets `spec.backgroundAudit == True`?  Could that case happen?

rfc/0009-audit-checks.md
```diff
@@ -35,67 +35,50 @@ As a Kubernetes developer, I want access to the audit checks results to allow me
...
+instead.
+
+It should implement a caching mechanism to avoid overloading the Kubernetes api server. We can use client-go caching.
+
+A background audit checks consist of the following steps:
 1. Get information about which policies have audit checks enabled
 2. Get the resources configured in the policies with audit checks enabled
-3. Send the resources that a policy is configured to validate to it and get the validation results.
+3. Request the policy evaluation for all resources to the `PolicyServer`
 4. Aggregate all the policies validation results
```
viccuad: I can't think of an easy to compile a list of resources that are being policed, to serve them to policies.

I think we should just compile a list of all resources. Then, as expected, create bogus requests for those resources, marking those requests as "audit" requests, and serving them directly to the PolicyServers instead of the Kube API. Then, receive all evaluations from their audit policies.
PolicyServers are the ones aware of which things can be evaluated: what kind of audit requests (not those for UPDATE ops for example) and audit policies (not those context-aware, or just looking for UPDATE ops for example).

rfc/0009-audit-checks.md
```diff
@@ -35,67 +35,50 @@ As a Kubernetes developer, I want access to the audit checks results to allow me
```
viccuad: I think we should add a user story:

```suggestion
As a Kubernetes developer, in the future, I want to be able to implement functionality in Kubewarden to add PolicyReportResults in real time for real requests and real policies.
```

I think we ought to make our future selves happier if we allow for Kubewarden to report all policy evaluation via PolicyReportResults in the future, not only audit ones. (This may mean re-evaluating the implementation approach).

rfc/0009-audit-checks.md
```diff
@@ -137,82 +120,52 @@ spec:
...
-to add the new fields used to store the audit checks resources and results.
-
-## Audit check results
+# Drawbacks
 
-Once the policy server has the audit checks results, it will populate the a [PolicyReport](https://github.com/kubernetes-sigs/wg-policy-prototypes/tree/master/policy-report).
-The report will contains all the success and failures  returned by the policies.
-Results from `ClusterAdmissionPolicy` (cluster wide policies) will be stored in the `ClusterPolicyReport`
-and the result from the `AdmissionPolicy` (namespaces policies) will be stored in the `PolicyReport`.
+Both Audit Background Scanner and Policy Server must write to the same PolicyReport. It might happen that both try to
```
viccuad:
```suggestion
Both Audit Background Scanner and PolicyServers must write to the same PolicyReport. It might happen that both try to
```

viccuad, line 133: We may have more than 1 PolicyServer trying to write. Hence I favour alternative A, Auditor Scanner having a rest API, and being the only one that can add PolicyReportResults.

rfc/0009-audit-checks.md
```diff
@@ -137,82 +120,52 @@ spec:
...
 
-# Alternatives
+If auditor is not available report data is lost. This data will be recovered in the next background scan.
 
-Additional to the `PolicyReport` report, policy server could add an annotation to the
-resources violating policies. However, this will require additional permissions to the policy server.
+Background Auditor Scanner is the only service writing `PolicyReports`, which makes easier to handle concurrent writes.
 
-Policies could have a dedicated function to evaluate resources in the audit checks
+## Run background scan in PolicyServer
```
viccuad:
```suggestion
## Run background scan in PolicyServers
```

(we can have several PolicyServers, we should update the document with that in mind).

rfc/0009-audit-checks.md
```diff
@@ -137,82 +120,52 @@ spec:
...
 
-If the number of resources for an individual policy is too high, the resources could
-be batched and multiple instances of the same policy could be run at the same time
-temporally to speed up the policy validations.  This horizontal scaling for all the
-policies under demand is a good topic for another RFC or idea discussion for the future.
+No need to create Background Auditor Scanner. Everything is done in `PolicyServer`. PolicyServer will take care 
+of doing the background scan.
 
+If background scan takes too long `PolicyServer` might be slower for evaluation requests, which would slow down creation
```
viccuad: (randomly opening thread here)

If I understand the problem correctly, we have:

1. Subset of resources that can be audited. Which?
2. Subset of policies that can be audited. Which? (those with `spec.backgroundAudit: true` only?)
3. Faked requests for audited requests.
    * Who creates them? (Background Audit Scanner?).
    * Are they to be sent against the Kube API? (if so, [using Kubernetes dry-run requests?](https://kubernetes.io/docs/reference/using-api/api-concepts/#dry-run) )
    * Or are they to be sent to PolicyServers directly?
4. Evaluation of faked requests. Performed by PolicyServers in parallel.
    * Are requests all sent and evaluated all at once, or can we create a queue of requests to be evaluated and evaluate them somewhat slowly to distribute the load over time?
    * If dry-run requests against Kube API, how does it work?
    * If instead requests are sent against PolicyServers, how do the PolicyServers know when to skip or evaluate an audit request with a specific policy? Didn't Kube API take care of this, by serving the request to the policy via webhook?
5. Each request returns a result. Save those results in PolicyReportResults, in parallel. How? Who does the write operation? How do we ensure it is thread-safe? Queues?
    * Would be nice if this would work for both faked audit requests or normal ones.

rfc/0009-audit-checks.md
raulcabello: I don't see a big problem with context-aware policies or those that make external calls. They would be evaluated with the context at this point in time, which give you an idea if this policy would reject a resource at this moment (it might change if the context changes).

Just CREATE requests will be audited, others operations should be ignored.

A operator can disable backgroundScan in any policy for cases like the [userinfo](https://github.com/kubewarden/rfc/pull/13#discussion_r965701469)

rfc/0009-audit-checks.md
raulcabello: >  PolicyServers are the ones aware of which things can be evaluated: what kind of audit requests (not those for UPDATE ops for example) and audit policies (not those context-aware, or just looking for UPDATE ops for example).

I don't believe PolicyServers are aware of which things can be evaluated. PolicyServers expose a endpoint per policy, then pass the request to the wasm module, however they don't know which resources can be evaluated. This information is inside the rules in the policy spec. That's why I think we should do the following steps:

1. Get information about which policies have audit checks enabled
2. Get the resources configured in the policies with audit checks enabled
3. Request the policy evaluation for all resources to the `PolicyServer`

rfc/0009-audit-checks.md
raulcabello: For example, if there is a policy that evaluates just UPDATE ops I think it should be ignored even if `spec.backgroundAudit == True`

rfc/0009-audit-checks.md
raulcabello: > Subset of resources that can be audited. Which?

I would say all. Any resource you think it can't be audited?

> Subset of policies that can be audited. Which? (those with spec.backgroundAudit: true only?)

yes, all policies with backgroundAudit: true

> Faked requests for audited requests.
    Who creates them? (Background Audit Scanner?).
    Are they to be sent against the Kube API? (if so, [using Kubernetes dry-run requests?](https://kubernetes.io/docs/reference/using-api/api-concepts/#dry-run) )
    Or are they to be sent to PolicyServers directly?

Fake requests are created by Background Audit Scanner and sent to relevant PolicyServer for evaluation, never to the kube API

> Evaluation of faked requests. Performed by PolicyServers in parallel.
    Are requests all sent and evaluated all at once, or can we create a queue of requests to be evaluated and evaluate them somewhat slowly to distribute the load over time?
    If dry-run requests against Kube API, how does it work?

That's a good point. PolicyServers will have a cache for all evaluations, which will reduce the number of evaluations performed. However we might want to send all evaluation for just one policy or one namespace before moving to the next one. What do you think?

>  If instead requests are sent against PolicyServers, how do the PolicyServers know when to skip or evaluate an audit request with a specific policy? Didn't Kube API take care of this, by serving the request to the policy via webhook?

I don't understand what you mean here. Can you explain it, please? 

>Each request returns a result. Save those results in PolicyReportResults, in parallel. How? Who does the write operation? How do we ensure it is thread-safe? Queues?
    Would be nice if this would work for both faked audit requests or normal ones

For fake requests as all requests happens in the Background Audit Scanner we can implement a sync mechanism using channel/mutex, or as previously mentioned we can wait until all request for a given namespace are finished, then update the PolicyReport. We will need to implement a retry mechanism if a different service update it at the same time. Imagine a PolicyServer write to a PolicyReport while a background scan is being done. This will give an error that the object was modified, and it should retry.

rfc/0009-audit-checks.md
raulcabello: Do you mean also adding failed evaluation to the PolicyReportResults?

rfc/0009-audit-checks.md
viccuad: Follow-up:
Kubernetes dry-run requests also help in evaluating requests correctly when they are mutated, as dry-run requests will get mutated by the mutating policy, and then validated by the rest of policies, closely simulating reality. If not using dry-run requests, policies would evaluate non-mutated requests when they would have been mutated.

viccuad: > I don't understand what you mean here. Can you explain it, please?

My musing is about, how does a PolicyServer know which Policy they need to use to evaluate a request? This was done previously by the webhook registration and Kube API server, we need to re-implement this. Moreover, how does it work for mutated requests? do we re-issue the mutated audit request for validating policies, as Kube API server does?

rfc/0009-audit-checks.md
viccuad: I don't intend to add it now, but to have an implementation that is compatible with such for future-proofing. I would expect the same mechanism for normal requests, just that the PolicyReportResults would be in a different new PolicyReport and ClusterPolicyReport.

Even if we don't implement this in the future, it could use the same http rest API to report results. The results could be consumed by open-telemetry, or via a UI for example.

rfc/0009-audit-checks.md
viccuad: Maybe I'm misunderstanding, as 2 and 3 seem exclusive to me (either approach 2 or 3).
I still can't think of an easy way to compile the list in 2. How do you obtain a list of resources for which a policy applies to?

Isn't it easier to do it the other way around, as in number 3, get the list of all resources, and naively try evaluation of their audit requests? This would mean to send all requests to all PolicyServers, because, how would one know which PolicyServer has which (Cluster)AdmissionPolicy that is relevant?

I would prefer to use dry-run and use the Kube API webhooks to take care of all of this.

rfc/0009-audit-checks.md
viccuad: What happens if somebody misconfigures a policy that is checking `userInfo`, and for that policy they set `spec.backgroundAudit == True`?

rfc/0009-audit-checks.md
raulcabello: You can see all resources that will be evaluated for a policy if you look at the `spec.rule` field. For example, in the [psp-capabilities](https://docs.kubewarden.io/quick-start) in the quick start just `Pods` will be evaluated.

Problems I see with kube api dry-run:
1. How do you know which policy rejected the request? We might be able to workaround this by looking at the error message, but doesn't seem like a nice approach.
2. It's not possible to know if a resource would be rejected for more than one policy, as it will be rejected for the one that first does the evaluation.
3. How do we skip evaluation for policies with background-scan=false? I don't see a way to do it with this.
4. What about policies in monitor mode? They will not reject any resources.

rfc/0009-audit-checks.md
raulcabello: Normal requests that were accepted will also be added to the PolicyReport, rejected requests for policies in monitor mode will also be added. Requests that were not accepted in enforce mode will not be added since these resources are not in the cluster. So why another PolicyReport and ClusterPolicyReport?

rfc/0009-audit-checks.md
raulcabello: We know which policy-server and which resources a policy will evaluate just by looking at the policy spec. We know the endpoint for each policy in the relevant policy-server because they always follow the same naming convention.
Mutation won't happen when running the background audit check. Background Auditor Check requests evaluation for a specific resource to a PolicyServer. If the response contains a mutation it will be ignored.

See [this](https://github.com/kubewarden/rfc/pull/10#discussion_r982073919) for dry-run

rfc/0009-audit-checks.md
raulcabello: wrong data might be added to the report. We should specify in the docs that `userInfo` will be different in the docs, and suggest to set `spec.backgroundAudit == false` for policies that check `userInfo`

rfc/0009-audit-checks.md
viccuad: > You can see all resources that will be evaluated for a policy if you look at the `spec.rule` field. For example, in the [psp-capabilities](https://docs.kubewarden.io/quick-start) in the quick start just `Pods` will be evaluated.

I suppose we would need to check also if the policy is namespaced via `AdmissionPolicy` or `namespaceSelector`.

> Problems I see with kube api dry-run:
> 
>     1. How do you know which policy rejected the request? We might be able to workaround
>         this by looking at the error message, but doesn't seem like a nice approach.

If the Audit Scanner is the one creating the dry-run requests, the PolicyServers could be the ones reporting back the results to the Audit Scanner, and they would know the policy.

>     2. It's not possible to know if a resource would be rejected for more than one policy, as it
>          will be rejected for the one that first does the evaluation.

This is a serious shortcoming of dry-run, and possibly a stopper.
 
>     3. How do we skip evaluation for policies with background-scan=false? I don't see a way
>         to do it with this.

I suppose that dry-run requests get marked somehow (I can only see the `?dryRun` in the query) and go through the admission phases, yet Kube API knows to now persist the results into etcd.  It may be the same for us?
Dry-run asks for [custom operators to mark themselves as `sideEffects` of `None` or `NoneOnDryRun`](https://kubernetes.io/docs/reference/using-api/api-concepts/#make-a-dry-run-request). This may tell that custom operators can know that a request if of type dry-run then.

>     4. What about policies in monitor mode? They will not reject any resources.

If the PolicyServers are the ones creating results (by creating PolicyReportResults or by hitting an Rest API), then they have this information.

rfc/0009-audit-checks.md
viccuad: Logging real requests being rejected into PolicyReports, or exposing that in an Rest API would be useful. Right now, one needs to go read the policy-server pod logs, and compile the tuple (PolicyServer, (Cluster)AdmissionPolicy, result), or deploy Jaeger.
Those results could be used in the UI for instantaneous feedback.

rfc/0009-audit-checks.md
viccuad: > We know which policy-server and which resources a policy will evaluate just by looking at the policy spec. We know the endpoint for each policy in the relevant policy-server because they always follow the same naming convention.

I suppose we need to look for namespaced `AdmissionPolicies`,  honour `spec.namespaceSelector`, `spec.objectSelector`, `spec.matchPolicy == equivalent` too. We ought to document this on the list of audited policies.

> Mutation won't happen when running the background audit check.

We ought to document this in the RFC, as part of the list of non-audited requests.

rfc/0009-audit-checks.md
viccuad: I just think it's a good idea to document specifically which type of requests and which type of policies take play in the audit feature.
Resolving this thread, as I have raised that point in others.

rfc/0009-audit-checks.md
```diff
@@ -0,0 +1,171 @@
...
+
+It should implement a caching mechanism to avoid overloading the Kubernetes api server. We can use client-go caching.
+
+A background audit checks consist of the following steps:
+1. Get information about which policies have audit checks enabled
+2. Get the resources configured in the policies with audit checks enabled
+3. Request the policy evaluation for all resources to the `PolicyServer`
+4. Aggregate all the policies validation results
+5. Publish a `PolicyReport` per namespace
+with the aggregated results and the `ClusterPolicyReport`.
```
viccuad: (As raised in [the other RFC](https://github.com/kubewarden/rfc/pull/13#discussion_r975215856))

- How atomically is a PolicyReport updated?
  - Does the PolicyReport contain all the evaluations for the *last run* of the audit of the namespace? (if results are present in the report, it is assumed that any result is the latest)
  - Or the *last run* of the resource? (if a result for resource is present in the report, it is assumed that the results are the latest for that specific resourse, but may not be for other resources)
- How frecuently are PolicyReportResults garbage-collected?
  - Do we only save the latest Result per (resource, policy) tuple?
  - Do we save older Results? How do we configure how many? on Kubewarden, or the PolicyReport itself?
    If we save older results, I fear it will not be efficient to try to find specific results for a resource.

rfc/0009-audit-checks.md
```diff
@@ -0,0 +1,171 @@
+
+|              |                                  |
+| :----------- | :------------------------------- |
+| Feature Name | Audit checks                           |
+| Start Date   | 08/05/2022                        |
+| Category     | enhancement,feature                 |
+| RFC PR       | https://github.com/kubewarden/rfc/pull/10  |
+| State        | **IN REVIEW**                     |
```
viccuad:
```suggestion
| State        | **ACCEPTED**                     |
```

Also, the filename is `0009-audit-checks.md`, and number 9 is already taken by the helm chart RFC (sorry).

rfc/0009-audit-checks.md
raulcabello: added in https://github.com/kubewarden/rfc/pull/10/commits/3172607122dadbd95a99e3433219816d2f3f9c59 
please let me know if you think there is still something missing

rfc/0009-audit-checks.md
raulcabello: I think this is something we should do in the future, but for now I would leave out of the scope of this RFC.

viccuad: Still, I'm just asking, if it's a good idea to provide an implementation that can support that use case in the future (like the user story).

rfc/0009-audit-checks.md
raulcabello: >How atomically is a PolicyReport updated? 

If both PolicyServer and Background Auditor Checks writes to the same PolicyReport we would need to implement some retry mechanism as it says [here](https://github.com/kubewarden/rfc/blob/audit-checks/rfc/0009-audit-checks.md?plain=1#L140-L142)

>How frecuently are PolicyReportResults garbage-collected? 

https://github.com/kubewarden/rfc/blob/audit-checks/rfc/0009-audit-checks.md?plain=1#L58-L59

rfc/0009-audit-checks.md
raulcabello: I have done some research on how dry-run works:
1. If you try to create an existing resource with dry-run, it will try to update the resource instead of creating. Which would result in wrong auditing data, since a policy evaluating creations wouldn't run. As a workaround we could change the name, that would trigger a creation event (however I don't like this)
2. AdmissionRequests has a dryRun parameter, so we know in the policy-server if the evaluation is a dry-run
3. There is no way to know which policy rejected a resource If we do the dry-run request from the Audit Scanner. The PolicyServer knows, then it would need to request the Audit Scanner to store the result. This is inconvenient because:
  - we need an extra call
  - what happens if Audit Scanner is not available? we lose data
  - there is no way to aggregate results. We would need to add PolicyReportResult one by one (inefficient).
  - when do we do the cleanup? when we start the background scanner? how do we know it has ended if request are make async from policy-server?
 4. We don't really care about the result of the dry-run call, since Audit Scanner would be waiting for PolicyServer requests. How do we know it has ended if request are make async from policy-server?

viccuad: > I have done some research on how dry-run works:
> 
>     1. If you try to create an existing resource with dry-run, it will try to update the resource instead of creating. 
>         Which would result in wrong auditing data, since a policy evaluating creations wouldn't run. As a
>         workaround we could change the name, that would trigger a creation event (however I don't like this)

Could it be that you are doing a `PUT` (create or update) instead of a `POST` (create)? If using `kubectl`, I suppose it does `PUT`.
I've tried it here using `kubectl create --raw` (that just takes the endpoint and does the REST API call), it works as expected and doesn't update the object:
```console
$ kubectl create --raw '/api/v1/namespaces/default/pods?dryRun=All' -f simple-pod.json
Error from server (AlreadyExists): pods "nginx" already exists
```
(tried the same with `kubectl proxy --port=8080 &` and a `curl -X POST`, but I'm failing to call it with the correct `?dryRun=All` query correctly)


> 3\. There is no way to know which policy rejected a resource If we do the dry-run request from the Audit Scanner. The PolicyServer knows, then it would need to request the Audit Scanner to store the result. This is inconvenient because:
> 
> 
>     * we need an extra call

I think we will need extra calls. We need to collect the results from all available PolicyServers. I'm not sure that naively retrying on all the PolicyServers is good enough. And I believe in the future we will need endpoints and a rest API anyways to expose realtime results on policy evaluation.

>     * what happens if Audit Scanner is not available? we lose data

If Audit Scanner is not available, we will not be creating the audit requests in the first place. And we will see that the Audit Scanner is not ready. I do like that if something breaks havoc in the cluster, I can start and stop the audit scanner, freeze the results in time by not letting accept more updates, etc. Even make it scalable, or sharded, by deploying Audit Scanners per namespace or so.

>     * there is no way to aggregate results. We would need to add PolicyReportResult one by one (inefficient).

We could implement a simple queue on the Audit Scanner (should be the default). We could use Kafka also if needed. In fact, if we see that we create too many PolicyReportResults because the cluster is busy, it could be configured to not persist them to etcd, but to serve them from a db. Not that we want to implement that in the first run, but the design allows for it.

>     * when do we do the cleanup? when we start the background scanner?

The cleanup is a duty of the Audit Scanner. The start of the background scan (creating the audit requests, and submitting them) is also duty of the Audit Scanner. Both could be configured as needed, be staggered, etc.

I think this simplifies saving and deleting PolicyReportResults. Instead of having X PolicyServers in charge of creating/deleting results from a single DB (PolicyReport), we can have 1 Audit Scanner per PolicyReport (or Audit Scanner queue/endpoint). This Audit Scanner is the sole owner of the creating/deleting results in the PolicyReport. Hence, we would have an Audit Scanner per namespace, matching their PolicyReport, and an Audit Scanner for the ClusterPolicyReport.

>  how do we know it has ended if request are make async from policy-server?

I understood it as the Audit Scanner being the ones creating and POSTing the requests, and waiting for the return code of the requests (409 conflict already exists). If all dry-run requests to the Kube API server gave a return, the audit run is complete. 

Alternatively, we could naively wait for PolicyServers to report back results. Each PolicyServer will return X results, one per policy that applies, for each 1 request. We have Y PolicyServers, so X*Y results per 1 requests.


----

Another point on dry-run:

5. Custom controllers need to behave correctly, don't generate side effects if run on dryRun, and explicilty advertise `sideEffects = None` or `NoneOnDryRun`. So far, I can't find one that doesn't do. Still they could be falsely self-advertising. In case they declare sideEffects, the request would not be created IIRC. Which means those resources would not be audited.

rfc/0009-audit-checks.md
viccuad: Thanks! what about `spec.objectSelector`, `spec.matchPolicy == equivalent`? Do we need to do something specific?
What if someone is using `objectSelector` in a policy, to only target objects with a `prod` label? We would running more policy evaluations than needed, and there would be results that shouldn't be there.

rfc/0009-audit-checks.md
viccuad: The approaches then contradict each other:
- 1. [Retry adding results to the PolicyReport](https://github.com/kubewarden/rfc/blob/audit-checks/rfc/0009-audit-checks.md?plain=1#L140-L142). So we are updating the PolicyReport all the time during the audit run.
- 2. [Wait until audit run is done to update the PolicyReport](https://github.com/kubewarden/rfc/blob/audit-checks/rfc/0009-audit-checks.md?plain=1#L58-L59).


> If both PolicyServer and Background Auditor Checks writes to the same PolicyReport we would need to implement some retry mechanism as it says [here](https://github.com/kubewarden/rfc/blob/audit-checks/rfc/0009-audit-checks.md?plain=1#L140-L142)

In my opinion we are wavering a bit the problem here. How do we implement this? naively retrying?  And what happens when there's a success in adding a new Result? The resource version changes, hence other retries may not be valid, and would need a GET to get the latest resource version. I think this can explode on us.
Also, I suppose we doing append-only on adding results, and if so, we should document it at so. And that deletion is a separate operation, that must be done at another time, and be thread safe.

> https://github.com/kubewarden/rfc/blob/audit-checks/rfc/0009-audit-checks.md?plain=1#L58-L59

But, right now we are saying that PolicyServers create/update/delete things in the PolicyReports. How does one PolicyServer know that the other PolicyServers have finished evaluating all requests of the current audit run? I don't think this is easily implemented, as the state would be implicitly spread between all PolicyServers.

Apart from the implementation, as a user and consumer of the PolicyReports,  knowing that the PolicyReport contains all the full evaluations for the *last audit run* is valuable to me. Plus, as Kubewarden devs, I think we can promise this.

raulcabello: PolicyServers just append new PolicyReports when evaluation happens (never in the background check). These evaluations will be removed when the background check happens, so after a background check we know that all resources have been fully scanned and the latest status of the cluster.

Background Auditor Check make all requests to PolicyServer, so it knows when all evaluation have finished and it then  updates the PolicyReport (it removes all previous records and updates with the latest cluster status).

I'm thinking now that it might be better to just do the background scan for now, and not store any evaluations in the PolicyServer. We would lose some of the real time info, but anyway the real time is not accurate because we are not tracking all events. I will add this and the dry-run as other options to the RFC.

rfc/0009-audit-checks.md
raulcabello: > Could it be that you are doing a PUT (create or update) instead of a POST (create)? If using kubectl, I suppose it does PUT.
> I've tried it here using kubectl create --raw (that just takes the endpoint and does the REST API call), it works as expected and doesn't update the object:

yes, you are right :) 

> I think we will need extra calls. We need to collect the results from all available PolicyServers. I'm not sure that naively retrying on all the PolicyServers is good enough. And I believe in the future we will need endpoints and a rest API anyways to expose realtime results on policy evaluation

That depends if we will want to store real time results directly from PolicyServer. Having a rest API have the problem of the auditor being offline and we lose data. Also having multiple auditors for HA is a challenge because we need to sync which one start the background check

> We could implement a simple queue on the Audit Scanner (should be the default). We could use Kafka also if needed. In fact, if we see that we create too many PolicyReportResults because the cluster is busy, it could be configured to not persist them to etcd, but to serve them from a db. Not that we want to implement that in the first run, but the design allows for it.

It is not straightforward to create a queue because everything is async and we don't know when it has finished. Communication is like Auditor -> kube api (using dry-run) -> (it might reach PolicyServer or not we don't know) -> back to Auditor (if  PolicyServer was reached). Kafka would be great, but I think it is way to big and uses a lot of resources.

> I think this simplifies saving and deleting PolicyReportResults. Instead of having X PolicyServers in charge of creating/deleting results from a single DB (PolicyReport), we can have 1 Audit Scanner per PolicyReport (or Audit Scanner queue/endpoint). This Audit Scanner is the sole owner of the creating/deleting results in the PolicyReport. Hence, we would have an Audit Scanner per namespace, matching their PolicyReport, and an Audit Scanner for the ClusterPolicyReport.

I think one audit scanner per namespace is too much, we can have one for all and then create queues or other sync mechanism per namespace

> I understood it as the Audit Scanner being the ones creating and POSTing the requests, and waiting for the return code of the requests (409 conflict already exists). If all dry-run requests to the Kube API server gave a return, the audit run is complete.
> 
> Alternatively, we could naively wait for PolicyServers to report back results. Each PolicyServer will return X results, one per policy that applies, for each 1 request. We have Y PolicyServers, so X*Y results per 1 requests.

we can wait until all kube API returns. However, PolicyServer **might not** return X results, one per policy that applies, for each 1 request. If another webhook rejects the request before the policy evaluation PolicyServer won't see any request. Someone might have [Pod Security Admission](https://kubernetes.io/docs/concepts/security/pod-security-admission/) deployed alongside Kubewarden, that would also interfere in the results

rfc/0009-audit-checks.md
```diff
@@ -0,0 +1,178 @@
...
+2. Get the resources configured in the policies with audit checks enabled. For `AdmissionPolicies` it will check resources
+within its namespace. For 'ClusterAdmissionPolicies' it will check in all the cluster or use the `namespaceSelector` filter
+if present.
+3. Request the policy evaluation for all resources to the `PolicyServer`
+4. Aggregate all the policies validation results
+5. Publish a `PolicyReport` per namespace with the aggregated results and the `ClusterPolicyReport`.
+
+In step 2, after fetching the resources for the first time,  they will be stored in a
+cache and used again in futures audit checks until the *time to live* period expired.
+Triggering another request to the Kubernetes API.
```
jvanz: This `TTL` should be a configuration.

rfc/0009-audit-checks.md
viccuad: Resolving, I consider it tackled with https://github.com/kubewarden/rfc/pull/10/commits/3172607122dadbd95a99e3433219816d2f3f9c59.

jvanz (commented):
I would say that the text looks good to me. But I need to think a little bit more about the comment in the thread  opened by Victor.

rfc/0009-audit-checks.md
flavio: > Normal requests that were accepted will also be added to the PolicyReport, rejected requests for policies in monitor mode will also be added. Requests that were not accepted in enforce mode will not be added since these resources are not in the cluster.

I totally agree with that.

@viccuad do you want to have rejected resources to show up into the `PolicyReport` and the `ClusterPolicyReport` too?

If that's what you want, I don't get it from the proposed change you made. I understand the proposed change as "as a kubernetes developer I want to be able to create/add entries to PolicyReport/ClusterPolicyReport", which is something I don't agree with. Creating/updating these resources should be done only by Kubewarden

flavio:
I've read through the comments, this is an overview of how I feel about this topic.

## Leveraging Kubernetes' dry-run feature

Using the dry-run feature of the API server was a clever idea, but I don't think this is going to work in our scenario.

Let's say we have two policies that are evaluating pods: `polA` and `polB`. Submitting a dry-run of an existing pod results in a failure. However, it would be impossible for us to know which policy actually rejected the pod. Maybe it was `polA`, maybe it was `polB`, maybe both of them are not fine with it. 

Because of that, I think the only way is to operate in a different way.

## The guiding principle

What's important for us is to come up with a solution that has a rock solid algorithm. We have to come up with a series of steps that receives some input data and produces output based on that. The last step of the workflow will lead to the creation of the `PolicyReport`. 

How the input data is obtained can be changed over the time. The goal is to have something we can ship and analyze against real world data.

Maybe in the beginning there are going to be limitations about what kind of scenarios we are going to support, but these should be overcome in the future without having to rewrite big parts of the code.

## Let's focus on Namespaces

I think Namespace are the minimal computational unit of the audit reports. We should define the steps needed to create the `PolicyReport` for one specific Namespace. Once this is done, the actual code can be run in parallel or in sequential mode against all the Namespaces, this doesn't matter too much.

Finally, the process to create the `PolicyReport` for a single Namespace, can be applied also to generate the `ClusterPolicyReport`, which is the one targeting only cluster-wide resources.

The following sections focus on each step of the process.

### Step 1: find relevant policies 

Given a Namespace to be audited, create a list of all the policies that are interested about the Namespace.

All the `AdmissionPolicy` resources that are defined inside of the namespace are part of the list.

For `ClusterAdmissionPolicy` resources, the Namespace selector has to be used to figure out if the audited Namespace is relevant.
  
> Note: we will consider only policies that inspect `CREATE` events

### Step 2: find the relevant Kubernetes resources

Now we have a list of all the policies that are interested about the Namespace being audited. We iterate over each one of them and create a dictionary that has:
* Key: the Kubernetes resource type, for example `Pod`
* Value: a list of policies that evaluate the given Kubernetes resource

The idea is to have a map of policies that tell us these information:

* `Pod` resources are relevant to `polA`, `polB`
* `Deployment` resources are relevant to `polA`
* `Ingress` resources are relevant to `polC`

### Step 3: fetch Kubernetes resources to be audited

Next we iterate over the keys of this dictionary and, for each one of them we query the API server to get all the resources of type `X` defined inside of our Namespace.

For example, given the previous example, we would end up with:

* List of all the `Pod` objects defined inside of the audited Namespace
* List of all the `Deployment` objects defined inside of the audited Namespace
* List of all the `Ingress` objects defined inside of the audited Namespace

**Note:** there's no need to do any caching of the response we get from the Kubernetes API server because of the following reasons:

* We work on a per Namespace basis, these query results are useful only in the context of the Namespace being audited
* The audit reports is generated every X minutes (maybe 10/30/60 minutes), cached data would be useless after this time

### Step 4: attempt to reuse previous evaluation results

Now we have a these information:

* List of policies interested by a certain type of Kubernetes resource, like `Pod`
* List of Kubernetes resources, `Pod` in this case, defined inside of the Namespace we are auditing

The code will then retrieve the `PolicyReport` for the namespace being audited.

We now iterate over each resource, for example over each `Pod` resource defined inside of the Namespace, and perform the following operations:

* Iterate over all the policies that are interested in `Pod` resources, for each tuple of (policy `X`, pod instance `Y`):
  * Look into the `PolicyReport` for an entry that features policy `X` evaluating pod instance `Y`:
    * If there's an entry, look at the evaluation timestamp of this entry. If policy update timestamp **and** pod instance update timestamp are older than the evaluation timestamp it means that nothing changed. Hence we are going to reuse the old evaluation result. Otherwise something changed, hence we need to perform an evaluation.
    * If there's no entry we will perform an evaluation

Once the outer loop is done (the one against all the Kubernetes resources defined inside of the audited Namespace that are affected by policies), our code will have all the data needed to create the `PolicyReport` for the namespace being audited.

### Step 4.5: perform evaluations

In the previous section, we saw how sometimes we need to perform a policy evaluation.

The evaluation happens by creating a fake `CREATE` event that has the Kubernetes object being audited as `request.object`.

As a first approach, I propose to perform this evaluation by making an HTTP request against the actual PolicyServer that is running the policy.

We can extend Policy Server to have a new endpoint for the audit checks. While the Kubernetes API server uses `/validate/POL_ID`, the audit scanner can file its requests against `/audit/POL_ID`. 

The `/audit` endpoint will be different from the `/validate` one in these regards:

* Generate different prometheus metrics: we don't want the overall `/validate` metrics to be polluted by the audit checks
* Generate different Open Telemetry traces: again, these should not be mixed with the `/validate` ones. We might even want to turn them off to make the audit execution faster
* Policies in `monitor` mode should still behave like the ones in `protect` mode. They have to produce rejections if needed

By issuing the audit request against a real PolicyServer, we solve these problems:

* Policies can be hosted on registries secured with credentials. We don't have to worry about being able to fetch the WebAssembly modules from these authenticated registries
* Policies might be hosted on insecure registires/registires with custom TLS certificates. Again, we won't have to bother about that
* Some data used by the policy (like external OCI manifests, sigstore verification results) can be cached by the PolicyServer and reused between the audit evaluations and the regular validations requests

Of course, sending all our auditing requests to the actual PolicyServer could lead to poor performance of the regular admission requests.
Currently, based on our load testing data, one single PolicyServer running with 2 workers can evaluate about 1600 requests per second. Based on this data, we think the requests generated by the audit checks should not clog the policy server.
If that turns out to be true, we can embrace different strategies to mitigate that.

Until we have real world data, it's not useful to eagerly optimize this aspect.

### Step 5: write `PolicyReport`

The code will overwrite the old `PolicyReport` instance with a new generated from the data kept in memory. By doing that, we don't have to worry about synchronizing the old contents with the new ones.

## Reducing the scope to keep things simple

In the beginning we should reduce the scope of the audit checks to be able to ship on time and iterate over the implementation by using real world data.

### No audit for policies targeting `*`

Certain policies can target any kind of Kubernetes resource. The safe annotations policy is an example of that.

Targeting all kind of resources would make more complicated to find all the Kubernetes resources contained inside of a Namespace that have to be audited.

To simplify things, we can state that policies targeting `*` are not going to be evaluated by the audit checks. Users have to be explicit about the Kubernetes resources a policy targets.

### No live update of PolicyResult objects

The proposed workflows requires that only one process is actively auditing the resources of a given Namespace.

That's because:

* The `PolicyResult` of the Namespace is used to reduce the number of auditing evaluations performed
* Once all the policies and involved resources are audited, the `PolicyReport` audit is going to be overwritten with the results of the audit

Users can still get live data about the rejections/approvals by looking both at Prometheus data and Open Telemetry traces.

We should focus on making the process of creating a `PolicyReport` as fast as possible. The reports are recreated on a regular basis. If a user wants a fresh `PolicyReport` without having to wait for the next scheduled run, we can provide a way for the user to trigger this operation.

Finally, if we really have to implement live update of `PolicyReport`, we can come up with an alternative architecture that will allow for that. For example, we could have all the write operations of `PolicyReport` to go through a dedicated "writer" service.

flavio:
Sorry for the long comment :sweat: . Let's talk more about that in a dedicated call instead of going through another round of comments. This is going to be faster

rfc/0009-audit-checks.md
viccuad: @flavio I would like to make the implementation compatible with adding real requests into `PolicyReport` and `ClusterPolicyReport` of their own.

Edited the suggestion for clarity.

rfc/0011-audit-checks.md
```diff
@@ -0,0 +1,354 @@
+
+|              |                                  |
+| :----------- | :------------------------------- |
+| Feature Name | Audit checks                           |
+| Start Date   | 08/05/2022                        |
+| Category     | enhancement,feature                 |
+| RFC PR       | https://github.com/kubewarden/rfc/pull/10  |
+| State        | **IN REVIEW**                     |
```
viccuad:
```suggestion
| State        | **ACCEPTED**                     |
```


Also, file should be `0012-audit-checks.md`, 11 is taken by the PolicyReport one.

rfc/0011-audit-checks.md
```diff
@@ -0,0 +1,354 @@
...
+
+
+# Summary
+
+The audit checks inspects the resources already deployed in the cluster and flags
```
viccuad:
```suggestion
The audit check inspects the resources already deployed in the cluster and flags
```

raulcabello:
The `objectSelector` is of type `metav1.LabelSelector` same as the `namespaceSelector`. That means we can use the same method [FormatLabelSelector](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/helpers.go#L169) to convert it to the type we need for doing the filtering. 

This changes the implementation proposed since we have to apply the `objectSelector` to get all pods that are relevant for a specific policy in a namespace. This is what we can do in order to avoid doing extra calls to the api server:

- Create the dictionary proposed in [step 2](https://github.com/kubewarden/rfc/pull/10#issuecomment-1265682099)
- Check if the resource matches the `objectSelector` before doing an evaluation of a policy. This can be done using the [Matches](https://github.com/kubernetes/apimachinery/blob/master/pkg/labels/selector.go#L51) method which doesn't make any call to the apiServer.

rfc/0011-audit-checks.md
```diff
@@ -0,0 +1,354 @@
...
+### Step 2: find relevant policies
+
+Given a Namespace to be audited, create a list of all the policies that are interested about the Namespace.
+
+All the `AdmissionPolicy` resources that are defined inside of the namespace are part of the list.
+
+For `ClusterAdmissionPolicy` resources, use the information stored inside of the
+dictionary that was created during the previous step.
+
+> Note: we will consider only policies that inspect `CREATE` events
```
viccuad, line 165: I would put this as part of the `### No audit for policies targeting *` paragraph. Listing the precondition of policies to be audited all in the same paragraph.

rfc/0011-audit-checks.md
```diff
@@ -0,0 +1,354 @@
...
+* List of all the `Ingress` objects defined inside of the audited Namespace
+
+**Note:** there's no need to do any caching of the response we get from the Kubernetes API server because of the following reasons:
+
+* We work on a per Namespace basis, these query results are useful only in the context of the Namespace being audited
+* The audit reports is generated every X minutes (maybe 10/30/60 minutes), cached data would be useless after this time
+
+### Step 5: attempt to reuse previous evaluation results
+
+Now we have a these information:
```
viccuad:
```suggestion
Now we have the following:
```

rfc/0011-audit-checks.md
```diff
@@ -0,0 +1,354 @@
...
+about the cluster-wide resources.
+
+Details about the `PolicyReport` format can be found inside of this [RFC](https://github.com/kubewarden/rfc/pull/13/files).
+
+## Reducing the scope to keep things simple
+
+In the beginning we should reduce the scope of the audit checks to be able to
+ship on time and iterate over the implementation by using real world data.
+
+### No audit for policies targeting `*`
```
viccuad, line 97: I would prefer these paragraph to be about all policies that are not auditable.
We should also mention policies that look at the user of the request.
It would be nice to mention the enabled-by-default `backgroundAudit` spec and metadata field to (Cluster)AdmissionPolicies  that marks them as auditable.

rfc/0011-audit-checks.md
```diff
@@ -0,0 +1,354 @@
...
+The `/audit` API will behave in a slightly different way compared to the current `/validate`
+endpoint. The differences have been illustrated by the previous section of things
+RFC.
+
+## Changes to ClusterAdmissionPolicy and AdmissionPolicy CRDs
+
+It's necessary a change in the policy CRDs to mark a policy to be run in the background audit checks.
+The field would be called `backgroundAudit` and its default value will be `true`.
+Therefore, all the policies will run in the audit checks by default, unless the user
+configure the other way. Example of policies CRDs with the `backgroundAudit` field:
```
viccuad, line 302: Are we gonna add the `backgroundAudit` field to metadata.yaml as we do with `mutating`?

rfc/0011-audit-checks.md
```diff
@@ -0,0 +1,354 @@
...
+
+No real time audit info are provided by this solution. Users have to rely on
+Prometheus metrics and Open Telemetry trace events to have live data coming
+from the cluster.
+
+We might want to address this limitation in the future by introducing a more
+complex architecture.
+
+`PolicyReports` size could be big if there are many resources and policies in the same namespace. More info on how big it 
+could be [here](https://github.com/kubewarden/rfc/pull/13#issuecomment-1253744208)
```
viccuad:
```suggestion
could be [here](https://github.com/kubewarden/rfc/pull/13#issuecomment-1253744208).
```

viccuad (commented):
Looks ready, just minor nitpicks here!

rfc/0011-audit-checks.md
```diff
@@ -0,0 +1,354 @@
...
+The following section describes how, given a Namespace, we will produce a `PolicyReport` describing the compliance status.
+
+The process to create the `PolicyReport` for a single Namespace can be applied also to generate the `ClusterPolicyReport`, which is the one targeting only cluster-wide resources.
+
+### Step 1: create association between `ClusterAdmissionPolicy` and namespaces
+
+`ClusterAdmissionPolicy` by default inspect events happening inside of all the namespaces, however users can limit their scope
+by using the `spec.namespaceSelector` field.
+
+We need to create a dictionary that holds the following information:
```
raulcabello, line 145: I would create a dictionary that holds the following information:
* key: namespace
* value: list of `ClusterAdmissionPolicies` relevant to this namespace
This would simplify step 3, because we can easily know all the `ClusterAdmissionPolicies` relevant to a namespace just by getting all values using the namespace as key.

rfc/0011-audit-checks.md
```diff
@@ -0,0 +1,354 @@
...
+
+* List of all the `Pod` objects defined inside of the audited Namespace
+* List of all the `Deployment` objects defined inside of the audited Namespace
+* List of all the `Ingress` objects defined inside of the audited Namespace
+
+**Note:** there's no need to do any caching of the response we get from the Kubernetes API server because of the following reasons:
+
+* We work on a per Namespace basis, these query results are useful only in the context of the Namespace being audited
+* The audit reports is generated every X minutes (maybe 10/30/60 minutes), cached data would be useless after this time
+
```
raulcabello, line 196: We need to take into account the `objectSelector` before starting with Step5.
The `objectSelector` is of type `metav1.LabelSelector` same as the `namespaceSelector`. That means we can use the same method [FormatLabelSelector](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/helpers.go#L169) to convert it to the type we need for doing the filtering.

We need to check if the resource matches the `objectSelector` before moving to step 5, ignore the resource if it doesn't match the `objectSelector`. This can be done using the [Matches](https://github.com/kubernetes/apimachinery/blob/master/pkg/labels/selector.go#L51) method which doesn't make any call to the apiServer.

raulcabello (approved):
LGTM, I just left a comment with a suggestion for the dictionary creation

jvanz (commented):
I cannot approve the PR because I'm the PR owner. But, besides other team mates comment, LGTM.

rfc/0009-audit-checks.md
flavio: The file has been renamed

rfc/0009-audit-checks.md
flavio: outdated, this comment was relevant with the the initial draft

rfc/0011-audit-checks.md
flavio: I've added a new section about that

rfc/0011-audit-checks.md
flavio: fixed

rfc/0011-audit-checks.md
flavio: This is now discussed in a broader way at the top of the RFC. I've added a new section

rfc/0011-audit-checks.md
flavio: The goal of this step is:

* Given a Namespace
* Given a list of Kubernetes resource types that are relevant to the policies to be evaluated,
* Get a list of all the Kubernetes instances defined inside of that Namespace

For example, you know you have 2 policies interested about `Pod` -> we get all the `Pod` instances defined inside of the Namespace. Once of the policies might have an `objectSelector`, but we don't care about that right now.

The filtering can be done during step 5, if this doesn't require extra calls to the API server. Which seems the case.
See below

rfc/0011-audit-checks.md
```diff
@@ -0,0 +1,354 @@
...
+      than the evaluation timestamp it means that nothing changed. Hence we are
+      going to reuse the old evaluation result. Otherwise something changed,
+      hence we **might** need to perform an evaluation.
+    * If there's no entry we **might** perform an evaluation
+  * If we reached this stage, it means the pod instance `Y` might need to be
+    evaluated by policy `X`. We have to check the `spec.ObjectSelector` field
+    defined inside of the policy (this applies both to `ClusterAdmissionPolicy`
+    and to `AdmissionPolicy`) and figure out if the `ObjectSelector` is
+    compatible with the pod instance `Y`. If that's true, we are going to
+    perform the actual evaluation
```
flavio, line 224: @raulcabello this is where the `objectSelector` is taken into account. IF and only IF we can determine whether pod instance `Y` is compatible with the given `objectSelector` **without** doing another API call.

If we have to perform an API call, then we should revisit the two steps and, for example, make multiple API requests to get the exact list of pod instances that are relevant to each policy

rfc/0011-audit-checks.md
flavio: I would add it. This is now mentioned in a new section I've just added

viccuad (approved):
LGTM! Thanks for the changes!

rfc/0011-audit-checks.md
raulcabello: perfect! I think we can use the [Matches](https://github.com/kubernetes/apimachinery/blob/master/pkg/labels/selector.go#L51) method without doing any extra API call

raulcabello (approved):
LGTM, thanks!

PR #12, opened 2022-08-31
Add RFC 10, Rancher integration of policies

### LINKED ISSUE #7: RFC: how to provide a list of policies to Rancher Manager
While working on the Rancher Manager integration, we need to find a way to show a list of the Kubewarden policies that exist.

Right now there's an hard-coded list of policies as a temporary workaround, but this is not a viable solution.

The data needed by Rancher manager is:

  * List of available policies
  * For each policy:
    * Name, version, description, validation/mutation, types of operations (CREATE, DELETE, ...), types of resources (Pod, Service, ...), ... basically all the information that is inside of the policy metadata
    * URL of the policy: where the policy is found
    * "Questions raw data": this is a yaml object that describes the settings a policy takes. Using this data the UI builds a tailor-made html form that is used by the user to configure the policy.
    * Signature status: is the policy signed or not

Right now the hard coded values can be found [here](https://github.com/jordojordo/kubewarden-charts/tree/main/charts).

## Acceptance criteria

* Write a RFC that shows possible solution(s) about how to make this data available to Rancher Manager
* The solution must take into consideration air-gapped environments
* Bonus: it would be nice to have a way to group related policies together. For example, have a "PSP best practices" group that allows users to easily find all the policies that are related to PSP. The same could be done for "Security best practices", which could include some PSPs, signature verification and some other conformance policies.

viccuad:
## Description

Relates to https://github.com/kubewarden/rfc/issues/7

[Rendered RFC](https://github.com/viccuad/rfc/blob/rancher-integration-policies/rfc/0010-rancher-integration-policies.md).

### DIFF
--- a/rfc/0010-rancher-integration-policies.md
+++ b/rfc/0010-rancher-integration-policies.md
@@ -48,7 +48,7 @@ Problems to solve:
 5. As a Kubewarden developer, I want to release a new Kubewarden chart/policy so
    it can be installed via Rancher Explorer UI.
 6. As a 3rd party developer, I want to release a policy on Artifact Hub so it can
-   be installed via Rancher Explorer UI.
+   be discovered and installed via Rancher Explorer UI.
 
 # Detailed design
 [design]: #detailed-design
--- a/rfc/0010-rancher-integration-policies.md
+++ b/rfc/0010-rancher-integration-policies.md
@@ -151,6 +151,14 @@ either under:
 It would then be retrieved with an OCI registry client. This would be more
 efficient than pulling full Wasm modules for all policies as in (A).
 
+This means policies would have the metadata saved in both:
+- Wasm module, as of today. Used by `kwctl`, `policy-server`.
+- OCI registry `manifest.config`.
+
+This would mean that policy-related metadata can be dropped from
+`policy/artifacthub-pkg.yml`, which should only contain things related to
+Artifact Hub.
+
 Airgap installations would consist on mirroring the OCI registry.
 Rancher airgap works by providing a list of images to be mirrored, and
 instructions for using a private registry. We would need to instruct Kubewarden
--- a/rfc/0010-rancher-integration-policies.md
+++ b/rfc/0010-rancher-integration-policies.md
@@ -178,3 +178,7 @@ with these formats, and can authenticate against private registries too.
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
+As in RFC-9, all needed images for the charts need to be mirrored in the Rancher
+DockerHub repository, instead of being consumed from upstream. Since policies
+are Wasm modules, we need to raise the request to provide a Rancher OCI
+registry, use Kubewarden's upstream one, or else.
--- a/rfc/0010-rancher-integration-policies.md
+++ b/rfc/0010-rancher-integration-policies.md
@@ -166,7 +166,7 @@ users to use an additional OCI private registry, and an OCI-capable client for
 performing the mirroring.
 
 A drawback could be finding frontend-capable OCI registry clients that can deal
-with these formats.
+with these formats, and can authenticate against private registries too.
 
 
 # Unresolved questions
--- a/rfc/0010-rancher-integration-policies.md
+++ b/rfc/0010-rancher-integration-policies.md
@@ -54,17 +54,27 @@ Problems to solve:
 Follow the process listed in [RFC-9, Rancher integration of Kubewarden
 charts](./0009-rancher-integration-charts.md).
 
-Carry the Helm chart code for the policy in a `rancher-chart` folder in the repo.
+Carry the Helm chart code for the policy in a `rancher-chart` folder in the
+policy repo.
 
-The resulting chart built from the policy repository will not be served and
-released by us, but used to build Rancher charts from source by using the
-[Package](https://github.com/rancher/charts/blob/dev-v2.6/docs/packages.md)
-format in https://github.com/rancher/charts.
-They will be submitted via a `packages/kubewarden/<policy>.package.yaml`, with
-`package.yaml::commit` pointing to the relevant `rancher-X` branch.
+Still, even if the chart will not be released to users, the repository can be
+consumed in Rancher UI (using git URL and branch) for testing the chart itself.
 
 For policy bundles, they can be in their own repository containing the chart
 and modifications (e.g: see `kubewarden-defaults` chart).
+
+The resulting policy chart built from the policy repository will not be served
+and released by us, but used to build Rancher charts from source by using the
+[Package](https://github.com/rancher/charts/blob/dev-v2.6/docs/packages.md)
+format in https://github.com/rancher/charts.
+Policy charts will be submitted via a
+`packages/kubewarden/<policy>.package.yaml`, with `package.yaml::commit`
+pointing to the relevant `rancher-X` branch. Rancher charts are build from those
+`package.yaml` files.
+
+A pull request bot needs to be created to make the submissions to either
+https://github.com/rancher/charts or our own staging fork, by bumping the
+`package.yaml::commit`.
 
 ## For 3rd party artifacts
 
@@ -89,10 +99,6 @@ they should be able to coexist.
 
 Needing new Helm charts to ship policies owned by the Kubewarden team.
 
-A pull request bot needs to be created to make the submissions to
-https://github.com/rancher/charts, and we will need external approvals for
-merging.
-
 ## For 3rd party policies
 
 Using the provided Artifact Hub API, given that it doesn't have pagination and
@@ -100,10 +106,10 @@ one needs to query for each package independently, will be inefficient. A cache
 could be implemented.
 
 The following metadata is not obtainable without reading the Wasm binary right
-right now, therefore should be added to `policy/artifacthub-pkg.yml`:
+now, therefore should be added to `policy/artifacthub-pkg.yml`:
 - Metadata not present in `policy/artifacthub-pkg.yml` yet present in
   `policy/metadata.yml`, namely the array of `rules`.
-- `question.yml` information. Not a problem per se, as 3rd party policies would
+- `questions.yml` information. Not a problem per se, as 3rd party policies would
   not need Rancher integration.
 
 Airgap installations need to catch the Artifact Hub metadata somehow.
@@ -172,5 +178,3 @@ with these formats, and can authenticate against private registries too.
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
-Alternative (B) looks the most promising, but more investigation on a
-frontend-capable OCI registry client needs to be done.
--- a/rfc/0010-rancher-integration-policies.md
+++ b/rfc/0010-rancher-integration-policies.md
@@ -39,8 +39,10 @@ Problems to solve:
 
 2. As a user, I want to install one Kubewarden policy owned by the
    Kubewarden/Rancher team via Rancher Explorer UI.
-3. As a user, I want to install several Kubewarden policies owned by the
-   Kubewarden/Rancher team via Rancher Explorer UI.
+3. As a user, I want to install a collection of Kubewarden policies owned by the
+   Kubewarden/Rancher team via the Rancher Explorer UI.
+4. As a Kubewarden developer, I want to allow users to deploy a collection of related policies
+   in a simple way. For example: "PSP best of", "Security best practices",...
 4. As a user, I want to install 3rd party Kubewarden policies via Rancher
    Explorer UI.
 5. As a Kubewarden developer, I want to release a new Kubewarden chart/policy so
--- a/rfc/0010-rancher-integration-policies.md
+++ b/rfc/0010-rancher-integration-policies.md
@@ -13,10 +13,8 @@
 This RFC explains how the Kubewarden policies, either owned by the Kubewarden
 team or 3rd party ones, can be integrated in Rancher Explorer.
 
-For policies owned by the Kubewarden team  it proposes a development workflow
-using git branches and automation. The needed Rancher Helm charts and Rancher
-Packages artifacts metadata is tracked in a separate orphan branch for each
-artifact's repo, and submitted to https://github.com/rancher/charts.
+For policies owned by the Kubewarden team it proposes building Rancher Helm charts
+that will be published under https://github.com/rancher/charts.
 
 For 3rd party policies not owned by the Kubewarden team,
 it proposes using the Artifact Hub API to poll for the metadata stored in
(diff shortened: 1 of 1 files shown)


rfc/0010-rancher-integration-policies.md
```diff
@@ -0,0 +1,162 @@
...
+- Obtain the list of available policies regardless of where they come from, and
+  their metadata to show in Rancher Explorer UI.
+- Allow for groups of policies to be installed.
+- Implement it efficiently and with low development maintenance costs.
+- Allow for airgap installations.
+
+## Examples / User Stories
+[examples]: #examples
+
+2. As a user, I want to install one Kubewarden policy owned by the
```
flavio: nitpicking, point number 1 is missing

rfc/0010-rancher-integration-policies.md
```diff
@@ -0,0 +1,162 @@
...
+   Kubewarden/Rancher team via Rancher Explorer UI.
+3. As a user, I want to install several Kubewarden policies owned by the
+   Kubewarden/Rancher team via Rancher Explorer UI.
```
flavio: This looks like something that could be done via the UI and a process of selection boxes, like with webmails where you select multiple messages and then do a "delete/archive'.

Maybe you mean something like:

```suggestion
3. As a user, I want to install a collection of Kubewarden policies owned by the
   Kubewarden/Rancher team via the Rancher Explorer UI.
4. As a Kubewarden developer, I want to allow users to deploy a collection of related policies
   in a simple way. For example: "PSP best of", "Security best practices",...
```

rfc/0010-rancher-integration-policies.md
```diff
@@ -0,0 +1,162 @@
...
+4. As a user, I want to install 3rd party Kubewarden policies via Rancher
+   Explorer UI.
+5. As a Kubewarden developer, I want to release a new Kubewarden chart/policy so
+   it can be installed via Rancher Explorer UI.
+6. As a 3rd party developer, I want to release a policy on Artifact Hub so it can
+   be installed via Rancher Explorer UI.
```
flavio:
```suggestion
6. As a 3rd party developer, I want to release a policy on Artifact Hub so it can
   be discovered and installed via Rancher Explorer UI.
```

rfc/0010-rancher-integration-policies.md
```diff
@@ -0,0 +1,162 @@
...
+## For policies owned by the Kubewarden team
+
+Follow the process listed in [RFC-9, Rancher integration of Kubewarden
+charts](./0009-rancher-integration-charts.md).
+
+Carry the Helm chart code for the policy in a `rancher-X` orphan branch, where `X`
+is the release tag of the policy being targeted.
+Using these branches allows for easy rebases and cherry-picks, and documentating
+needed changes per commit. This also separates the Rancher vendored code, which
+may be of no interest to policy authors, and may coexist with other vendors.
```
flavio: Wouldn't be simpler to just have it under a dedicated directory (like `rancher/`)?

Is the dedicated git branch an orphan one (meaning completely disconnected from `main`, à la `github-pages` so to speak)

rfc/0010-rancher-integration-policies.md
```diff
@@ -0,0 +1,162 @@
...
+where one can set the `repository kind` to `Kubewarden policies`. With this call
+one can obtain the list of Kubewarden policies. One would then query for each of
+the policies with [GET
+/packages/kubewarden/{repoName}/{packageName}](https://artifacthub.io/docs/api/#/Packages/getKubewardenPoliciesDetails),
+and obtain its `name`, `version`, `signed` status, and `data/kubewarden-*`
+information.
+
+Artifacts owned by the Kubewarden team are also present in Artifact Hub, hence
+to avoid listing them twice, they should be skipped if `provider` of the package
+is `kubewarden`, and/or the URL starts with `https://github.com/kubewarden/`.
```
flavio: I think this is a decision that has to be taken later on, when implementing the UI. We might want to not filter the duplicates, we should instead make clear the provenance (artifact hub vs rancher helm charts repo) because this has an impact on supportability (I'm thinking long term).

For example, Artifact Hub might have a more recent version compared to the one inside of Rancher's helm chart repository.

rfc/0010-rancher-integration-policies.md
```diff
@@ -0,0 +1,162 @@
...
+- Or in `manifest.config` (https://github.com/opencontainers/artifacts/blob/main/artifact-authors.md#optional-defining-config-schema)
+
+It would then be retrieved with an OCI registry client. This would be more
+efficient than pulling full Wasm modules for all policies as in (A).
+
+Airgap installations would consist on mirroring the registry.
+
+A drawback could be finding OCI registry clients that can deal with these
+formats.
+
```
flavio, line 180: I think the main drawback is that this solutions requires more code to be written into Rancher Manager and potentially extending the airgap mechanism of Rancher manager.

rfc/0010-rancher-integration-policies.md
```diff
@@ -0,0 +1,162 @@
...
+## For 3rd party policies
+
+Using the provided Artifact Hub API, given that it doesn't have pagination and
+one needs to query for each package independently, will be inefficient. A cache
+could be implemented.
+
+Obtaining the policies via the API is still missing:
+- Metadata not present in `policy/artifacthub-pkg.yml` yet present in
+  `policy/metadata.yml`, namely the array of `rules`.
```
flavio, line 111: If we **really** want to provide a better integration with Rancher Manager, we could add all these information into the custom annotations that `artifacthub-pkg.yml` has. We could even provide a tool that takes `metadata.yml` and updates an existing `artifacthub-pkg.yml` file.

flavio (commented):
Great analysis :clap: 

I propose a variation on the alternative approach which is an hybrid between the core proposal and alternative "A":

* Store the additional information inside of the Wasm metadata
* Have a tool that reads the embedded metadata and produces as output all the Rancher helm chart files

The rest would work like the core proposal. What do you think?

viccuad:
As an example of a Helm chart for a policy, see https://github.com/kubewarden/allow-privilege-escalation-psp-policy/pull/30.

This example is implemented directly in the `main` branch of the policy repo, instead of an orphan branch, and is reusing its readme, license, and files.

rfc/0010-rancher-integration-policies.md
viccuad: A folder could be fine.
One could put the needed Helm files in root too, and use `.helmignore` to filter the rest. This allows for reusing readme, license, and so.
But, if we want the UI to take care of the Helm Chart, then the chart should be `catalog.cattle.io/hidden: true` so we shouldn't need to care about anything of that, no need to show the readme, and all config is done via questions.yaml.

See https://github.com/kubewarden/allow-privilege-escalation-psp-policy/pull/30

flavio:
> I propose a variation on the alternative approach which is an hybrid between the core proposal and alternative "A":
> 
> * Store the additional information inside of the Wasm metadata
> 
> * Have a tool that reads the embedded metadata and produces as output all the Rancher helm chart files

Looking at the contents created via https://github.com/kubewarden/allow-privilege-escalation-psp-policy/pull/30, I think the only reasonable thing to include inside of the policy metadata would be the `questions.yml` stuff. I wonder if this is worth the effort...

rfc/0010-rancher-integration-policies.md
viccuad: Good point, amended.

rfc/0010-rancher-integration-policies.md
viccuad: expanded in https://github.com/kubewarden/rfc/pull/12/commits/2745bd85485273610b26877d3556def055ed7799.

rfc/0010-rancher-integration-policies.md
viccuad: True, it's the obvious play. I should have been more verbose there, fixed in last version.

viccuad:
Committed clarification of points, and reworded to use folder instead of branch.

> Looking at the contents created via https://github.com/kubewarden/allow-privilege-escalation-psp-policy/pull/30, I think the only reasonable thing to include inside of the policy metadata would be the questions.yml stuff. I wonder if this is worth the effort...

It also provides a simple airgap story for the metadata (download Helm charts, mirror OCI wasm modules).

I still favour Alternative B (use OCI's `manifest.config`). To me, feels like the correct implementation. It's a field that is present for normal container images, simplifies airgap, removes unneeded metadata from artifacthub-pkg.yml.

We just need a client that can pull the manifest of an artifact from an OCI registry (e.g: 30 secs search, https://github.com/Pixeladed/oci-registry-js), and to do so on `kwctl push` too. We aren't making much of a new promise, just substituting storing the metadata in artifacthub/helm chart for the OCI repo.

rfc/0010-rancher-integration-policies.md
```diff
@@ -0,0 +1,176 @@
...
+
+A drawback could be finding frontend-capable OCI registry clients that can deal
+with these formats, and can authenticate against private registries too.
+
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+Alternative (B) looks the most promising, but more investigation on a
+frontend-capable OCI registry client needs to be done.
```
flavio: I think this can be removed now, we agreed to proceed with the approach proposed by the core proposal

rfc/0010-rancher-integration-policies.md
```diff
@@ -0,0 +1,176 @@
...
+charts](./0009-rancher-integration-charts.md).
+
+Carry the Helm chart code for the policy in a `rancher-chart` folder in the repo.
+
+The resulting chart built from the policy repository will not be served and
+released by us, but used to build Rancher charts from source by using the
+[Package](https://github.com/rancher/charts/blob/dev-v2.6/docs/packages.md)
+format in https://github.com/rancher/charts.
+They will be submitted via a `packages/kubewarden/<policy>.package.yaml`, with
+`package.yaml::commit` pointing to the relevant `rancher-X` branch.
```
flavio: Does it mean we will have our own fork of https://github.com/rancher/charts (maybe as a Kubewarden org)? I will assume we do.

In addition to the current tasks, before tagging a new policy release we will ensure the contents of the `rancher-chart` directory are uptodate. Then we will tag the new version.

After having done that, we will go to our fork of rancher charts and:
  * Create a branch for updating this policy
  * Copy the contents of `rancher-chart` from the policy into the right directory
  * Create a PR

Is that correct?

rfc/0010-rancher-integration-policies.md
viccuad: Yes, will update and remove that.

rfc/0010-rancher-integration-policies.md
flavio: Clarified over a call

rfc/0010-rancher-integration-policies.md
viccuad: Done in https://github.com/kubewarden/rfc/pull/12/commits/fcf6b07eb44c2f0d9b9ce08fefb1f4e896b7bafd.

rfc/0010-rancher-integration-policies.md
viccuad: (as talked on call)

More or less. Submission into the Rancher charts repository happens by providing a [package.yaml](https://github.com/rancher/charts/blob/dev-v2.6/docs/packages.md), which accepts [several ways to point to an upstream chart](https://github.com/rancher/charts/blob/dev-v2.6/docs/packages.md#upstreamoptions). Rancher charts are built from the package definitions.

I would point to upstream policy charts using the github repo, subdirectory. We could have a bot that bumps those in a fork of https://github.com/rancher/charts.

Added explanation on https://github.com/kubewarden/rfc/pull/12/commits/fcf6b07eb44c2f0d9b9ce08fefb1f4e896b7bafd.

rfc/0010-rancher-integration-policies.md
```diff
@@ -0,0 +1,188 @@
...
+## For policies owned by the Kubewarden team
+
+Follow the process listed in [RFC-9, Rancher integration of Kubewarden
+charts](./0009-rancher-integration-charts.md).
+
+Carry the Helm chart code for the policy in a `rancher-chart` folder in the
+policy repo.
+
+Still, even if the chart will not be released to users, the repository can be
+consumed in Rancher UI (using git URL and branch) for testing the chart itself.
```
flavio: Did you have time to look into the option we were talking: adding these policies helm charts to the https://github.com/kubewarden/helm-charts repository?

raulcabello (commented):
Thanks @viccuad ! I don't think it is a good idea to place kubewarden policies in https://github.com/rancher/charts 

In my opinion we should leave https://github.com/rancher/charts just for apps that can be deployed in your cluster (kubewarden for example)

We could create a separate chart repository for the policies something like https://github.com/rancher/kubewarden-policies-charts. Drawback is that users would need to add it manually, or would it possible to add this new repository when kubewarden is installed? (I don't think that's possible, but I'm not sure..)

rfc/0010-rancher-integration-policies.md
```diff
@@ -0,0 +1,188 @@
...
+This alternative provides the same implementation for both kubewarden-owner
+policies, Rancher-mirrored policies, and 3rd party policies. Just serve each of
+them on their respective OCI registries.
+
+A drawback could be finding frontend-capable OCI registry clients that can deal
+with these formats, and can authenticate against private registries too.
+
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
```
viccuad, line 183: (randomly starting thread here)

> In my opinion we should leave https://github.com/rancher/charts just for apps that can be deployed in your cluster (kubewarden for example)
>
> We could create a separate chart repository for the policies something like https://github.com/rancher/kubewarden-policies-charts. Drawback is that users would need to add it manually, or would it possible to add this new repository when kubewarden is installed? (I don't think that's possible, but I'm not sure..)

Submitting the Kubewarden policies into the Rancher distribution means that they will be supported in Rancher, which is what we want. They should all depend on the specific charts (ideally, just 1, that installs and configures the whole Kubewarden stack).

rfc/0010-rancher-integration-policies.md
viccuad: Yes, I'm iterating on it. I have opened https://github.com/kubewarden/helm-charts/pull/130/files.

Argh, sorry, amended the intro in the past, but not the rest of the document. Updated it in https://github.com/kubewarden/rfc/pull/12/commits/f703d00905b4540e295e2f222be4012c0804e54d

rfc/0010-rancher-integration-policies.md
raulcabello: Policies will appear as apps in the Rancher App Catalog. I think this might be confusing

viccuad: You have a point there. Maybe the RFC will need to be amended in the future to submit the policies to a different repo. For now, I'm happy as it is, I'm more concerned with proving is viable by crafting the Helm charts and questions.yaml.

raulcabello: yes, I agree. I'm happy as it is for now too :smile:

raulcabello (approved):
LGTM. We might need to make some modifications in the future based on the feedback we get for the questions.yaml and where to place the policies. I'm happy merging this now. 
Thanks for the hard work @viccuad!

jvanz (approved):
LGTM! :D
Thank you @viccuad for this work!

viccuad:
Merging! we can make adjustments later on.

PR #13, opened 2022-09-07
Add RFC about policy-report

flavio:
Add an RFC that describes the format used to store our policy reports.

This is kinda of a split of https://github.com/kubewarden/rfc/pull/10


rfc/0009-policy-report.md
```diff
@@ -0,0 +1,323 @@
...
+* How many policy evaluations led to a resource being accepted, rejected or
+  could not be performed because of an error.
+  Policy failures happen when an error occurs during the policy evaluation,
+  for example, because of a runtime panic
+* Detailed results about all the evaluations performed. An evaluation result contains
+  the following information: the policy being used, the resource that was
+  evaluated, the final outcome (accept/reject/failure) and the eventual message
+  returned by the policy
+
+The `PolicyReport` of the Namespace is kept updated during the lifetime of the
```
raulcabello: What if we don't keep updating the `PolicyReport` of the namespace? When a new audit check is triggered we could replace the old `PolicyReport` with the new one. 
If we keep adding new evaluations to the same `PolicyReport` I'm a bit concern we might end up with a big `PolicyReport` stored in etcd (it has limitations as you already mentioned).  It would also simplify the code.

We can assume a `PolicyReport` for a namespace is something that happened at some point with the state of the cluster at this time. And keep always the report for the latest check

We could keep some old reports instead of deleting them

rfc/0009-policy-report.md
```diff
@@ -0,0 +1,323 @@
...
+scans. Hence they will never show inside of the reports.
+
+# Drawbacks
+
+We will have just one `PolicyReport` per Namespace. Depending on the number of
+policies and Kubernetes resources defined inside of the Namespace, the size of
+the `policyReport.spec.results` could be significant.
+
+This could lead to `PolicyReport` objects too bit to be stored and/or degrade
+cluster performance because of the load put on ETCD.
```
raulcabello: Maybe add an option to not include evaluation that passed in the results? I think having information about why an evaluation failed is more important.

raulcabello (commented):
This looks great! Thanks @flavio! I left some ideas

rfc/0009-policy-report.md
```diff
@@ -0,0 +1,323 @@
...
+section about `PolicyReport`.
+
+## Changes to `ClusterAdmissionPolicy` and `AdmissionPolicy`
+
+We will extend the `ClusterAdmissionPolicy` and `AdmissionPolicy` resources to have
+a new optional field called `background`.
+This field is going to hold a boolean value, which is going to be set to be `true` by default.
+
+Policies that have `background` set to false will be ignored during the background
+scans. Hence they will never show inside of the reports.
```
viccuad: I would expect to have `PolicyReports` information for non-background policy runs too, be it enforcing or monitoring, to be fair. 
It's a bit weird that getting a rejection from an enforicing policy on a new CREATE request may or may not end in the `PolicyReports`, or may need to wait for the audit background job to run (and still, it will not be about the same request).

Maybe having `[Cluster]PolicyReports` for enforcing/monitoring policies, and`[Cluster]PolicyReports` for auditing.

rfc/0009-policy-report.md
```diff
@@ -0,0 +1,323 @@
...
+  the following information: the policy being used, the resource that was
+  evaluated, the final outcome (accept/reject/failure) and the eventual message
+  returned by the policy
+
+The `PolicyReport` of the Namespace is kept updated during the lifetime of the
+cluster, meaning:
+
+* New Policies might become "interested" about the events happening inside of
+  the cluster. New evaluations results will be added to the PolicyReport
+* Existing Kubernetes objects and policies can change over the time: some evaluations
```
viccuad:
```suggestion
* Existing Kubernetes objects and policies can change over time: some evaluations
```

rfc/0009-policy-report.md
viccuad: I like this idea. It makes also trivial to know if the ns/cluster is sane; if there's an empty "audit" `PolicyReport` (or one without errors/failures), all is fine. If updated checks are yet to be run, one sees the PolicyReport from the previous successful run with old `PolicyReportResult` timestamps. Antiquated PolicyReports could be archived somehow if desired.  

I would expect to have another set of `PolicyReports` and `ClusterPolicyReports`, that would contain only results from "monitor/enforcing" policies, and new requests. These "monitor/enforcing" reports would provide instant feedback if a policy is rejecting or approving requests, and these could contain an array of `PolicyReportResults` that is queue of reports, and that is garbage collected slow enough that the state of those resources would be present on the latest "audit" `PolicyReport`.

rfc/0009-policy-report.md
flavio: There are certain types of policies that we cannot audit in a reliable way. For example, a policy that relies on the `userInfo` section of a [request](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#request) cannot be audited.
That's because the `userInfo` "faked" by the audit is not going to be relevant.

There might be other corner cases. An operator should have the option to exclude a policy from the background scans.

rfc/0009-policy-report.md
flavio: It's up to us to decide how we are going to leverage the `PolicyReports` CR. However, other tools made for kyverno will take certain assumptions, if we do not respect these assumptions the results might be odd.

The main assumption is that a `PolicyReport` contains information about about all the evaluations done against the resources that were inside of a Namespace at a certain time. It is not tracking only the failures. By tracking also successful evaluations two main goals are achieved:

  * As a user I know with absolute certainty the compliance status of all the resources inside of a Namespace. If the PolicyReport doesn't contain any information about k8s resource X, that means the background checks has not evaluated it yet.
  * As Kubewarden authors, our code will know policy `polX` is interested about k8s reource `podA`. The code can look into the `PolicyReport` and look for an entry about `polX` evaluating `podA`. If no entry is found, the evaluation takes place. If the entry is found, we can look at the evaluation timestamp and compare it against the modification timestamps of `polX` and `podA`; if nothing changed since the last evaluation result we can just skip it and save time.

So the `PolicyReport` will contain only relevant information inside of it:
  * evaluations that are still valid, Old evaluation results are kept only when the policy AND the resource have not been changed since the evaluation time. This is done to reduce the number of evaluations we have to perform.
  * evaluations about resources/policies that have been deleted are removed

Despite that, I think there's the risk of having a big fat `PolicyReport` when a Namespace is crowded by k8s resources and/or many policies are enforced.

flavio:
There's one thing I forgot to mention. We are going to introduce an audit report feature to Kubewarden that is going to use the contents of this RFC. A background scan is going to be run at certain intervals and this is going to update `PolicyReport` objects.

However, there's also something else we can do in the future that leverages these CRDs. We could for example add entries to the `PolicyReport` in a live fashion to keep track of resources that have been rejected by our admission controller. This is something that kyverno does. We **might** want to do that too

rfc/0009-policy-report.md
flavio: See https://github.com/kubewarden/rfc/pull/13#discussion_r965715042 - this addresses the comment

rfc/0009-policy-report.md
```diff
@@ -0,0 +1,323 @@
...
+`AdmissionPolicy` and `ClusterAdmissionPolicy`.
+
+Only the most recent audit runs are kept. There's also a limit on the number of
+detailed reports added to a policy. This is done to prevent the clogging of ETCD,
+as described in the "Drawbacks" section above.
+
+The main limitation about this approach is not going to be able to reuse the
+tool that is going to be built around the `PolicyReport` CRDs.
+
+# Unresolved questions
```
viccuad: Tangential; for non-audit policy reporting (not the case in this RFC), if we have mutating policies, is there a possibility that several policies mutate a resource consecutively, getting trapped in an endless loop of mutations?

For example, these 2 policies may trigger each other in an endless loop.
- Policy A: mutates, acts on update, adds `annot: prod` if not present.
- Policy B: mutates, acts on update, if `annot: prod` present, change it to `annot: prod-touched`.

This may bring DOS, or crash of the cluster if the PolicyReports grow too large.

rfc/0009-policy-report.md
raulcabello: Those are good points. I like the idea of not redoing evalutation for resources that haven't changed. 
I was just trying to think of ways to reduce the size of the `PolicyReport`. However it will probably just be ok since other tools are already using this approach and it looks like this is not a big issue.

It would be better to stick with the same approach, so we can use tools that are already made for `PolicyReport` (e.g. policy reporter)

rfc/0009-policy-report.md
```diff
@@ -0,0 +1,323 @@
+
+|              |                                  |
+| :----------- | :------------------------------- |
+| Feature Name | Policy Report                           |
```
viccuad:
```suggestion
| Feature Name | Policy Report  for Audit background checks                         |
```
Or similar. I just want to make clear that it is about the background checks that simulate requests instead of logging real requests.

rfc/0009-policy-report.md
viccuad: That's a good point. I think it deserves to be mentioned in the RFC. Better still if we document the list of request types that will not be faked/audited, to provide to the users. Maybe that would be present in the upcoming audit feature RFC?

rfc/0009-policy-report.md
```diff
@@ -0,0 +1,323 @@
...
+| Start Date   | 08/05/2022                        |
+| Category     | enhancement,feature                 |
+| RFC PR       | https://github.com/kubewarden/rfc/pull/13  |
+| State        | **IN REVIEW**                     |
```
viccuad:
```suggestion
| State        | **ACCEPTED**                     |
```

rfc/0009-policy-report.md
```diff
@@ -0,0 +1,323 @@
...
+  for example, because of a runtime panic
+* Detailed results about all the evaluations performed. An evaluation result contains
+  the following information: the policy being used, the resource that was
+  evaluated, the final outcome (accept/reject/failure) and the eventual message
+  returned by the policy
+
+The `PolicyReport` of the Namespace is kept updated during the lifetime of the
+cluster, meaning:
+
+* New Policies might become "interested" about the events happening inside of
```
viccuad: Policies may be written to check UPDATE operations of PolicyReport. This will trigger an endless loop of entries to the PolicyReport, and possible DOS of the policy server/cluster. Do we need to care about this? Should we disallow policies from checking wgpolicyk8s.io/v1alpha2 CRDs (that includes PolicyReports, PolicyReportResults..).

rfc/0009-policy-report.md
```diff
@@ -0,0 +1,323 @@
...
+* `timestamp`: the time the result was found
+* `result`: this is a string enum. These are the values we will be using:
+  `pass`, `fail` and `error`
+* `resources`: this is a list of `core/v1.ObjectReference` objects. We will
+  put only one object reference inside of this list. This is going to be a
+  reference to the resource that has been analyzed by the policy
+* `resourceSelector`: we are not going to use it, because we are using the `resources` field
+* `message`: this is used when the policy result is `fail` or `error`. It will hold
+  the output message provided by the policy
+* `properties`: we are not going to use this free-form dictionary in the beginning
```
viccuad, line 171: I wonder if we should add versioning for our usage of PolicyReports. And if so, if it could go here with a `version: v1`.

rfc/0009-policy-report.md
```diff
@@ -0,0 +1,323 @@
...
+
+All the results are going to be collected into the `PolicyReport` object defined
+inside of the Namespace. There's going to be only one `PolicyReport` per Namespace.
+
+The `PolicyReport` resource has the following fields:
+
+* `apiVersion` and `kind`: these are defined by the CRDs
+* `metadata`: this is the usual `meta/v1.ObjectMeta` resource. In the
+  beginning we will set only the `name` field. We will do that by using the
+  following pattern: `polr-ns-<namespace name>`
```
viccuad: I wonder if we should add versioning to our specific usage of PolicyReports. If so, maybe a `metadata.annotation.version == v1`?

rfc/0009-policy-report.md
```diff
@@ -0,0 +1,323 @@
...
+
+## Namespace report
+
+A Namespace can be "affected" both by `ClusterAdmissionPolicy` and by `AdmissionPolicy` policies.
+When auditing a Namespace, only some Kubewarden policies are going to be interested
+about the events happening inside of that Namespace. In addition to that,
+the same Kubernetes resource could be evaluated by multiple policies. This will
+lead to several policy evaluations being performed.
+
+The `PolicyReport` is a namespaced CR that provides these information:
```
viccuad: Things that I find missing:
- How atomically is a PolicyReport updated?
  - Does the PolicyReport contain all the evaluations for the *last run* of the audit of the namespace? (if results are present in the report, it is assumed that any result is the latest)
  - Or the *last run* of the resource? (if a result for resource is present in the report, it is assumed that the results are the latest for that specific resourse, but may not be for other resources)
- How frecuently are PolicyReportResults garbage-collected?
  - Do we only save the latest Result per (resource, policy) tuple?
  - Do we save older Results? How do we configure how many? on Kubewarden, or the PolicyReport itself?
    If we save older results, I fear it will not be efficient to try to find specific results for a resource.

rfc/0009-policy-report.md
```diff
@@ -0,0 +1,323 @@
...
+* `source`: identifier for the policy engine that manages this report. We will
+  set that to be `kubewarden`
+* `policy`: this is the name of the policy that evaluated the resource. This
+  could be both a `ClusterAdmissionPolicy` or a `AdmissionPolicy`.
+  Because of that we cannot use the name of the policy, there could be clashes
+  between cluster and namespaced policy names. We will instead use the following
+  convention to build the contents of this field: `<[cap|ap]>-<policy name>`.
+  For example, a `ClusterAdmissionPolicy` named `provileged-containers` will
+  be identified via `cap-privileged-containers`. On the other hand, a `AdmissionPolicy`
+  called `required-labels` will be identified via `ap-required-labels`.
```
jvanz: Don't we [add prefixes](https://github.com/kubewarden/kubewarden-controller/blob/1c8afcce20f11f1de5056b8adcda53b51e455b3d/apis/policies/v1/clusteradmissionpolicy_types.go#L175) in the cluster policies to avoid this kind of issue? I think we do not need this in the reports.

rfc/0009-policy-report.md
```diff
@@ -0,0 +1,323 @@
...
+## `PolicyReport` detailed overview
+
+For each Namespace, Kubewarden will find all the policies that are interested
+about actions happening inside of it. This is going to be a mix of
+some `ClusterAdmissionPolicy` policies and all the `AdmissionPolicy` policies
+that are defined inside of the Namespace.
+
+Next, this list of policies is going to be filtered: only policies interested
+in `CREATE` events are going to be considered.
+The audit report is not going to simulate `DELETE` and `UPDATE` events.
```
jvanz: Won't we miss important evaluations when using only CREATE policies?

I'm trying to image an scenario where an user have a policy to check important resource configuration to avoid service disruption if the resource misses some important info during an update (e.g. environment variable, container settings, etc). In this scenario, our audit report will not report issues and our users will not be notified upfront.

rfc/0009-policy-report.md
jvanz: Do we have some stats from CNCF to let us know the average size of the namespaces? I'm wondering how big is the namespaces. Maybe this etcd size constraint is not a big problem now and we can iterate over it in the future. 

I see we wondering if the size of the report will be an issue. For that, I think we need to add more data in the discussion. Like, what's the limit evaluations number in the reports to get a good performance? In other words, how many result can we add to the report before seeing a performance degradation? Is there an easy way to check this?

rfc/0009-policy-report.md
jvanz: Cannot we use this field to add relevant information about the policy settings? Like if it is in `protect` or `monitor` mode.

rfc/0009-policy-report.md
viccuad: I think this is an internal string in the controller, and never gets set in the resource itself. But at least, we could adopt the same convention with `clusterwide-`.

rfc/0009-policy-report.md
viccuad: If that policy checks for CREATE, UPDATE, then we will simulate creating that resource that is missing the environment variable, and all is fine. If it's only UPDATE, it will not get checked by the continous audit feature and it will not go into the  PolicyReport because of it.

But it will get picked by the usual behaviour of policies when the real UPDATE operation happens. And then, we need to decide if we want to also log real request outcomes in a separate PolicyReport.

rfc/0009-policy-report.md
raulcabello: I think we add the prefix `clusterwide-` or `namespaced-nsname- `to the webhook configutarion and the name in the policies configmap to avoid having a `ClusterAdmissionPolicy` and an `AdmissionPolicy` with the same name.
I would use the same naming convention for `PolicyReports`

rfc/0009-policy-report.md
viccuad: As I understand it, the PolicyReports specified in this RFC cover only the continuous audit feature, so if there's any `fail` policy result on the results, it means that resource is violating policies, and is already in the cluster. It is equivalent to running all current policies in `monitor` mode against exisiting resources (as they may have changed with time).

rfc/0009-policy-report.md
raulcabello: we should definitely think about how to keep the PolicyReport up to date. This is a complex topic that might be worth to cover in a different RFC as pointed out [in the unresolved questions](https://github.com/kubewarden/rfc/pull/13/files#diff-8540a44feec39379c6de0278dc397e7fbd508815de206a73d8d86a12f9978138R322)

rfc/0009-policy-report.md
jvanz: Yeah, I made a confusion with the webhook and policy names. Sorry. We need the prefix indeed.

jvanz:
Just sharing a thought that I'm not sure if it fits in some thread already open.

I've reviewed the last [CNCF survey](https://www.cncf.io/reports/cncf-annual-survey-2021/) trying to find some additional information about cluster sizes. Trying to figure out if the report size could be a future problem. It has information about cluster sizes in terms of containers and machines running. In the survey we can see that 73.59% of the interviewees run clusters with 999 containers maximum. Let's assume for the sake of the discussion that each of the container has a resource witch is verified by all the policies (e.g. pod, deployment, etc). So, in the worst scenario, if all the policies are clusterwide, the number of entries in the report will be `999 * number of the policies`. Our recommended policy list in the `kubewarden-defaults` chart has 6 policies. Which led us to 5994 possible evaluations. Is it too much? I do not know.

In the other hand, the scenario above is unlikely to happen. Because in the same survey we learn that most of organizations separate their teams and applications by namespaces or separate clusters. So, the predominant case will be multiple smaller reports.

rfc/0009-policy-report.md
viccuad: [raised in the other RFC](https://github.com/kubewarden/rfc/pull/10#discussion_r981237592), resolving this.

rfc/0009-policy-report.md
flavio: All of that should be handed by another RFC

flavio:
> In the other hand, the scenario above is unlikely to happen. Because in the same survey we learn that most of organizations separate their teams and applications by namespaces or separate clusters. So, the predominant case will be multiple smaller reports.

Thanks for sharing this survey data. I agree with you, I doubt all the 999 containers are going to be located into 1~5 namespaces. I've looked into the survey data, the majority of users is relying on `Namespace`s to separate both applications and teams. Hence, I think we will end up with lots of smaller/medium sized reports

rfc/0009-policy-report.md
flavio: We need a predictable way to generate the payload used to simulate the event. This can be done only for `CREATE` and `DELETE` operations. `UPDATE` operations are hard to simulate, each policy has its own internal logic that can lead to different outcomes based on what we give as an input.

In the beginning I would start with just focusing on `CREATE` operations, later we might want to expand to cover also `DELETE` ops.

rfc/0009-policy-report.md
flavio: Done

rfc/0009-policy-report.md
flavio: This is really a corner case, I would not focus on that.

I think that, given `PolicyReport` are namespaced and contain security-related details, regular users should not be allowed to DELETE/UPDATE them. They should just have READ access to them. This can be done via RBAC policies. We might want to document that.

flavio: I've extended the RFC to point this out

rfc/0009-policy-report.md
flavio: Do you think we should do that because we might change the way we "consume" the `PolicyReport` document in the future?

flavio:
I've made a round of final adjustments. I think we can go ahead and merge the PR once the last open question is addressed.

rfc/0009-policy-report.md
viccuad: More that I think we may extend our usage by adding `PolicyReportResult` `properties` and so on. Both have admittedly low probabilities.
I'm happy without versioning them, don't intend to block on it.

viccuad (approved):

PR #15, opened 2023-01-18
First iteration of the metadata RFC

flavio:
How to automate the creation of `metadata.yml` and `artifacthub-pkg.yml`.

### DIFF
--- a/rfc/0014-metadata.md
+++ b/rfc/0014-metadata.md
@@ -1,0 +1,616 @@
+|              |                                  |
+| :----------- | :------------------------------- |
+| Feature Name | Unified metadata                 |
+| Start Date   | Jan 18 2023                      |
+| Category     | [Category]                       |
+| RFC PR       | [fill this in after opening PR]  |
+| State        | **ACCEPTED**                         |
+
+
+# Summary
+[summary]: #summary
+
+Policies have metadata information that must be made accessible to different
+consumers (policy-evalutor, artifacthub). Each consumer ingests the metadata
+in slightly different formats. We want to provide a way to reduce the manual
+operations required to keep the different metadata files up to date.
+
+# Motivation
+[motivation]: #motivation
+
+Currently each Kubewarden policy has a `metadata.yml` file that defines some
+metadata information about the policy. This information is used by policy-evaluator
+(hence by kwctl and policy-server) when executing a policy. It's mandatory for
+policies to be "annotated" with this metadata in order to be used.
+
+Policies published on [Artifact Hub](https://artifacthub.io) have an additional
+metadata file, called `artifacthub-pkg.yml`.
+
+There's some duplication of data between the two files, keeping these two files 
+in sync is error prone and time expensive. There's the risk of having "metadata drift"
+between the two files.
+
+This proposal aims to provide a solution to this problem.
+
+## Examples / User Stories
+[examples]: #examples
+
+## Publishing a policy on Artifact Hub for the first time
+
+As a policy author, I want to publish my policy on Artifact Hub. My policy has
+already a `metadata.yml` file, however I have to copy-paste, adapt and expand
+the contents of `metadata.yml` into the `artifacthub-pkg.yml` file.
+
+I would rather have a way to automatically generate the `artifacthub-pkg.yml` file.
+
+## Tag a new release of a policy
+
+As a policy author, I'm about to publish a new version of my policy. This policy
+is also registered on Artifact Hub.
+
+I would like to have some automation that updates the contents of the already
+existing `artifacthub-pkg.yml` file.
+
+# Detailed design
+[design]: #detailed-design
+
+Each Kubewarden policy must have a `metadata.yml` file. While only certain ones
+are going to have a `artifacthub-pkg.yml` file.
+
+We want to have the following workflow:
+
+* User creates a `metadata.yml` file. This file is scaffolded when the policy
+  is created for the 1st time
+* User enters proper data inside of `metadata.yml`
+* User keeps annotating the policy in the usual way: `kwctl annotate --metadata-path `metadata.yml`
+
+If the user wants to publish the policy on Artifact Hub, we will offer the following
+workflow:
+
+* User extends the `metadata.yml` file with some annotations that are specific
+  to Artifact Hub
+* User invokes: `kwctl scaffold artifacthub --metadata-path metadata.yml`.
+  This command will generate the `artifacthub-pkg.yml` file
+
+The freshly created `kwctl scaffold artifacthub` can be run as many times
+as the user wants. The contents of `artifacthub-pkg.yml` are going to be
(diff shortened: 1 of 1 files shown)


rfc/0014-metadata.md
```diff
@@ -0,0 +1,616 @@
...
+
+```yaml
+annotations:
+  kubewarden/resources: Pod, Deployment
+```
+This is going to be created by copying the value of this `metadata.yml` annotation:
+
+```yaml
+annotations:
+  io.artifacthub.resources: Pod, Deployment
```
viccuad, line 439: this is a new `metadata.yml` annotation, duplicating `rules.resources` yet in uppercase, singular, etc, correct?

rfc/0014-metadata.md
```diff
@@ -0,0 +1,616 @@
...
+  io.kubewarden.policy.url: https://github.com/kubewarden/verify-image-signatures
+  io.kubewarden.policy.source: https://github.com/kubewarden/verify-image-signatures
+  io.kubewarden.policy.license: Apache-2.0
+  io.kubewarden.policy.usage: |
+    How to use the policy
+```
+
+The `artifacthub-pkg.yml` file of the same policy has the following contents:
+
+```yaml
```
viccuad, line 121:
```suggestion
```yaml
---
# artifacthub-pkg.yml
```

rfc/0014-metadata.md
```diff
@@ -0,0 +1,616 @@
...
+It's important to point out that a user can completely avoid this workflow and
+keep managing the `artifacthub-pkg.yml` file as he prefers.
+
+## Comparing `metadata.yml` and `artifacthub-pkg.yml`
+
+The following snippet is taken from one of our policies, it
+includes all the mandatory fields:
+
+```yaml
+rules:
```
viccuad, line 90:
```suggestion
---
# metadata.yml
rules:
```

rfc/0014-metadata.md
```diff
@@ -0,0 +1,616 @@
+|              |                                  |
+| :----------- | :------------------------------- |
+| Feature Name | Unified metadata                 |
+| Start Date   | Jan 18 2023                      |
+| Category     | [Category]                       |
+| RFC PR       | [fill this in after opening PR]  |
```
viccuad:
```suggestion
| RFC PR       | https://github.com/kubewarden/rfc/pull/15  |
```

rfc/0014-metadata.md
```diff
@@ -0,0 +1,616 @@
...
+We thought about using JSON Schema to describe the format of our metadata. This is interesting, it could provide
+value to our users. However it doesn't address the problem of automating the generation of
+`artifacthub-pkg.yml` and `metadata.yml`.
+
+We can revisit this topic later on.
+
+
+# Unresolved questions
+
+Should we extend the `metadata.yml` file or should we create an "umbrella" configuration file?
```
viccuad, line 620: For the policy writer, my vote goes to extending the `metadata.yml` with optional keys as specified in this rfc, instead of an umbrella jsonnet. I find extending `metadata.yml` simpler, doesn't need learning jsonnet, and provides 1 fewer level of indirection that the author needs to learn about. Can be summarized with "If you want to publish to artifacthub, add this optional fields, run kwctl scaffold, and submit there". The only hiccup is that the annotations are only a string value, but I think is worth it to deserialize twice for them or something similar.

For the implementation details, I'm still unsure if using jrsonnet & static-compiling with musl is feasible. I will have a look at it.

rfc/0014-metadata.md
```diff
@@ -0,0 +1,616 @@
...
+There are however some recommended fields:
+
+```yaml
+annotations:
+  io.kubewarden.policy.title: verify-image-signatures
+  io.kubewarden.policy.description: A Kubewarden Policy that verifies all the signatures of the container images referenced by a Pod
+  io.kubewarden.policy.author: "Raul Cabello Martin <raul.cabello@suse.com>, Victor Cuadrado Juan <vcuadradojuan@suse.de>"
+  io.kubewarden.policy.url: https://github.com/kubewarden/verify-image-signatures
+  io.kubewarden.policy.source: https://github.com/kubewarden/verify-image-signatures
+  io.kubewarden.policy.license: Apache-2.0
```
viccuad, line 114: as pointed out in https://github.com/kubewarden/echo/issues/11#issuecomment-1397346695,
we could add a `rancher/hide-ui` annotation, analogous to [the one used for rancher charts](https://github.com/kubewarden/helm-charts/blob/main/charts/kubewarden-crds/Chart.yaml#L33).

rfc/0014-metadata.md
flavio: yes

rfc/0014-metadata.md
jvanz: Even if I think that jsonnet is cool. I don't see no big advantages to adopt now. Our needs seem to simple to worth adding another lib with the burden to learn how to maintain this new file format. In my humble opinion our needs are quite simpler. We just need to generate the artifact hub from our pre existing files. As we already have `kwctl` to do that hard work. Thus, I would say we can extend the `metadata.yml` file for now.

jvanz: I'm wondering if we cannot transform the `metadata.yml` file into the "umbrella" configuration file. The first idea that come to my mind after reading the question was to move the data from the annotations in the `metadata.yml` file into actual fields in the same file. Maybe using JSON format and missing the comments (no `jsonnet` at this point). With that, we have the type checks for free.  `kwctl` should be able to translate this new "umbrella" `metadata.yml` into anything (e.g. the annotation for the policy wasm and artifact hub file) with no big pain to the policy writers, isn't it? If we do that, it can be a first step to use the `jsonnet` in the future (if we decided to use it at all)

But AFAICS, that will require changes in the UI as well. Therefore, if this end to be a good idea, we may need to postpone it

viccuad: I agree with José.

For following the proposed path, I have created the outline of the cards in https://github.com/kubewarden/kubewarden-controller/issues/389, I hope it's ok. Edits/comments/etc welcome!

viccuad (commented):
To me, this RFC is ready besides the `rancher/hide-ui` annotation.

rfc/0014-metadata.md
flavio: Yes, but I would not mention that explicitly inside of the RFC.

All the annotations that being with `io.artifacthub.<something>: <value>` would be translated to

```yaml
annotations:
  <something>: <value>
```

inside of the `artifacthub-pkg.yml` file.

jvanz (approved):
LGTM!

rfc/0014-metadata.md
viccuad: fair enough, resolving!

viccuad (approved):
LGTM!

PR #16, opened 2023-02-14
Add RFC about context aware policies

flavio:
Kubewarden has already support for context aware policies, but this is kinda of a POC.

This RFC provides a more structured approach to solve this problem.

### DIFF
--- a/rfc/0015-context-aware-policies.md
+++ b/rfc/0015-context-aware-policies.md
@@ -1,0 +1,244 @@
+|              |                                  |
+| :----------- | :------------------------------- |
+| Feature Name | Context Aware Policies           |
+| Start Date   | 13-02-2023                       |
+| Category     | [Category]                       |
+| RFC PR       | [fill this in after opening PR]  |
+| State        | **ACCEPTED**                         |
+
+
+# Summary
+[summary]: #summary
+
+Define how policies can obtain additional information from Kubernetes API server
+at evaluation time.
+
+# Motivation
+[motivation]: #motivation
+
+In certain cases a policy requires additional information about the state of the
+Kubernetes cluster in order to take a validation decision.
+These kind of policies are called "context aware". 
+
+Kubewarden must allow policy authors to write context aware policies. At
+the same time, we must prevent rogue policies to perform abuses against the
+cluster.
+
+## Examples / User Stories
+[examples]: #examples
+
+> As a Policy author,
+> I want to obtain a list of all the other Ingress objects
+> defined inside of the cluster in order to accept or reject the
+> creation of another Ingress object.
+
+> As a Kubernetes administrator,
+> I want to know what kind of Kubernetes resources a policy can
+> query at evaluation time.
+
+> As a Kubernetes administrator,
+> I want context-aware policies to be able to access only the list
+> of Kubernetes resources that the policy author declared.
+> Attempts to access other types of resources should be blocked and
+> reported to me.
+
+# Detailed design
+[design]: #detailed-design
+
+Kubewarden currently features a prototype implementation of context aware policies.
+This implementation is limited and this RFC plans to entirely replace it.
+
+Currently we are not aware of any policy making use of this implementation.
+
+## Policy metadata
+
+Context aware policies must mention explicitly the types of
+Kubernetes resources they are going to access at evaluation time.
+
+This is required to provide transparency to the Kubernetes administrator who
+will deploy the policy.
+
+Policy author must mention the Kubernetes resources inside of the policy
+metadata:
+
+```yaml
+contextAware:
+# - group: <group>
+#   version: <version>
+#   resource: <resource>
+- version: v1
+  resource: pods
+- group: apps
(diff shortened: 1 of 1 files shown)


rfc/0015-context-aware-policies.md
```diff
@@ -0,0 +1,244 @@
+|              |                                  |
+| :----------- | :------------------------------- |
+| Feature Name | Context Aware Policies           |
+| Start Date   | 13-02-2023                       |
+| Category     | [Category]                       |
+| RFC PR       | [fill this in after opening PR]  |
```
viccuad:
```suggestion
| RFC PR       | https://github.com/kubewarden/rfc/pull/16  |
```

rfc/0015-context-aware-policies.md
```diff
@@ -0,0 +1,244 @@
...
+
+# Detailed design
+[design]: #detailed-design
+
+Kubewarden currently features a prototype implementation of context aware policies.
+This implementation is limited and this RFC plans to entirely replace it.
+
+Currently we are not aware of any policy making use of this implementation.
+
+## Policy metadata
```
viccuad, line 53: Questions,
- As I understand, the policy metadata is not used by policy-server to allow or deny requests, and is only provided as a courtesy for Kubernetes administrators, correct?
  (I'm concerned on trusting policy metadata, as a malicious policy author could annotate the policy as they see fit, granting themselves the permissions to read Secrets in the cluster for example).
- The ClusterAdmissionPolicy definition is the one that requests the contextAware capabilities, and they are accessible only if that ClusterAdmissionPolicy is scheduled in a PolicyServer that has contextAware enabled, correct?

Maybe it's worth it to signify even more obviously that the policy metadata is indeed just a courtesy and not taken into account when scheduling the policy, but I can't come up with a better UX.

rfc/0015-context-aware-policies.md
```diff
@@ -0,0 +1,244 @@
...
+  leak sensitive information to other policies. For example, if a trusted policy
+  requires access to Secret resources, all the other hosted policies will
+  automatically gain access to this type of resources too.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+This RFC does not cover the following details:
+
+* The waPC protocol used to exchange information between the guest and the hosted
```
viccuad:
```suggestion
* The waPC protocol used to exchange information between the guest and the host
```

rfc/0015-context-aware-policies.md
```diff
@@ -0,0 +1,244 @@
...
+* The caching mechanism used by Policy Server to store these information
```
viccuad:
```suggestion
* The caching mechanism used by Policy Server to store this information
```

rfc/0015-context-aware-policies.md
flavio: >     * As I understand, the policy metadata is not used by policy-server to allow or deny requests, and is only provided as a courtesy for Kubernetes administrators, correct?
>       (I'm concerned on trusting policy metadata, as a malicious policy author could annotate the policy as they see fit, granting themselves the permissions to read Secrets in the cluster for example).

The policy metadata is filled by the developer and contains the information about which resources the policy is going to access.
This is then used by the Kubernetes administrator to create the allow list inside of `ClusterAdmissionPolicy` definition.

An attacker could say his policy needs access to `Pod` using the metadata information, but behind the scene he can try to access also `Secret` objects.
However, the Kubernetes administrator will compose the allow list by looking at what the policy declares inside of its metadata. Hence, when the policy will attempt to access a `Secret` this request is going to be denied by Policy Server and the admin will be notified about that.

If the attacker wants to access `Secret` objects he must be explicit about that inside of the Policy metadata.
It's up to the Kubernetes admin to carefully review what the policy declares under the `contextAware` section, and question whether access to a certain resource is actually needed.

To make an analogy, it's like when reviewing the permissions of a "torch" application inside of the Android/Apple store. If the "torch" app requires access to your contact list or microphone you know there's something wrong going on.


>     * The ClusterAdmissionPolicy definition is the one that requests the contextAware capabilities, and they are accessible only if that ClusterAdmissionPolicy is scheduled in a PolicyServer that has contextAware enabled, correct?

All the Policy Servers are capable of exposing Kubernetes resources to any kind of policy. That happens as long as the policy is requiring access to one of the resources that are mentioned inside of `policies.yml` file for **that** policy.

Does that clarify your questions?

viccuad: Yep! perfect :). Resolving.

viccuad (approved):
LGTM! I really like the per-policy allowlist.

jvanz (approved):
LGTM! :D

flavio:
I did a minor change to the configuration format of the `contextAware` value, see https://github.com/kubewarden/rfc/pull/16/commits/41723ce982405a67a56b23a1c71d38a884891ea4

I would merge this proposal, if that's fine with you

viccuad (approved):

PR #17, opened 2023-02-16
Small change to context aware policies RFC

flavio:
Current policies have already a metadata attribute named `contextAware` which has a boolean value.

The initial proposal wanted to reuse this attribute to store a list of kubernetes resources.

Unfortunately this would be too complicated to do. We would have to write some custom deserialization code. Because of that, the RFC has been updated to use a different name for this metadata attribute.

### DIFF
--- a/rfc/0015-context-aware-policies.md
+++ b/rfc/0015-context-aware-policies.md
@@ -62,7 +62,7 @@ Policy author must mention the Kubernetes resources inside of the policy
 metadata:
 
 ```yaml
-contextAware:
+contextAwareResources:
 # - apiVersion: <apiVersion>
 #   kind: <kind>
 - apiVersion: v1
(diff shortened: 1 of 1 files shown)


flavio:
I just realized, should we use `contextAwareResources` everywhere, also inside of the `ClusterAdmissionPolicy` CRD and inside of Policy Server configuration file?

I think it would make everything consistent. What do you think about that?

viccuad:
I'm ok with using `contextAwareResources` everywhere including CRDs, and deprecating `contextAware`. It's a pity that we need to bump the CRDs, but it's needed.
I suppose we should have a new CRD version, deprecate the old version, and provide a conversion to the new one (that just sets contextAwareResources empty?)

jvanz:
> I just realized, should we use `contextAwareResources` everywhere, also inside of the `ClusterAdmissionPolicy` CRD and inside of Policy Server configuration file?
> 
> I think it would make everything consistent. What do you think about that?

Fine by me.

flavio:
> I'm ok with using `contextAwareResources` everywhere including CRDs, and deprecating `contextAware`. It's a pity that we need to bump the CRDs, but it's needed. I suppose we should have a new CRD version, deprecate the old version, and provide a conversion to the new one (that just sets contextAwareResources empty?)

We never had `contextAware` inside of the CRDs, we only had that inside of the policy metadata

flavio:
I've updated the PR to use `contextAwareResources` everywhere. Please double check :)

viccuad (approved):

PR #21, opened 2023-06-28
Add RFC about WASM cache

flavio:
Initial version of the RFC.

### DIFF
--- a/rfc/0016-wasm-cache.md
+++ b/rfc/0016-wasm-cache.md
@@ -1,0 +1,212 @@
+|              |                                  |
+| :----------- | :------------------------------- |
+| Feature Name | WASM cache                       |
+| Start Date   | Jun 28 2023                      |
+| Category     | [Category]                       |
+| RFC PR       | [fill this in after opening PR]  |
+| State        | **ACCEPTED**                         |
+
+
+# Summary
+[summary]: #summary
+
+This RFC aims to provide a way to reduce the startup time of Policy Server.
+
+# Motivation
+[motivation]: #motivation
+
+Policy Server is restarted every time its configuration changes. This includes
+changes done to the policies that are deployed through it. This includes adding
+or removing policies, plus any kind of change done to an already deployed one
+(like a different configuration value, or a version bump).
+
+During the startup Policy Server performs these time expensive operations:
+
+* Download all the policies
+* Optimize the WASM code for execution
+
+These operations are repeated every time, because each the state of each Policy
+Server instance is not preserved across restarts.
+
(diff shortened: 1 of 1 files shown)


flavio:
This is a draft, I didn't have time to proofread it since I have to leave for holidays.

@jvanz / @viccuad: feel free to take a look at that if you have time. As you will see I think there are other approaches that might be easier to accomplish. We can discuss more once I'm back

viccuad:
LGTM here, the listed alternatives are well documented. From the listed ones, I favour an optional PV and no StatefulSets. I look forward to follow-up discussions.

rfc/0016-wasm-cache.md
```diff
@@ -0,0 +1,226 @@
...
+it and then, before starting to listen for incoming requests, push the pre-built
+module to the OCI registry.
+
+Multiple Policy Server instances could be optimizing and pushing the same WASM
+module at the same time. However, the OCI registry can handle these concurrent
+push requests.
+
+## The caching OCI registry
+
+The OCI registry used for caching would not be exposed outside of the cluster.
```
viccuad, line 146: in case the cluster is airgapped, we could reuse the OCI registry inside of the airgap.

jvanz (commented):
I'm okay with the proposed approach. But I just had an idea...

What if do we remove the load of pre compiling the policies from the policy server? We could use a job which will download the policy pre-compile it and store in a PV. Therefore, policy servers does not write to it just read it. No write concurrent access. The controller can watch for new policies, triggers this job, and only after the successful job execution update/restart the policy server. This job could use a new `kwctl` command to optimize the policies and write the result into the PV read after by the policy server.

flavio:
> What if do we remove the load of pre compiling the policies from the policy server? We could use a job which will download the policy pre-compile it and store in a PV. Therefore, policy servers does not write to it just read it. No write concurrent access. The controller can watch for new policies, triggers this job, and only after the successful job execution update/restart the policy server. This job could use a new `kwctl` command to optimize the policies and write the result into the PV read after by the policy server.

Thanks for the suggestion. I still prefer one of the alternative approaches because this would not complicate the rollout strategy used by the controller to start a new instance of the policy server.

Let's experiment a bit with the kubernetes leases and a shared PV. I'll keep you posted.

flavio:
Closing the RFC as rejected. The alternative solutions illustrated inside of the RFC are more appealing.

I'm going to open a new RFC to explain how one of the alternative solutions can be adopted to solve this problem

flavio:
Sorry, I accidentally closed the PR. I'll instead merge it, as it's supposed to be. The RFC has status `REJECTED`

PR #22, opened 2023-07-18
New RFC: Policy Server - reuse modules

flavio:
This RFC is a follow up from the now rejected [RFC 16](https://github.com/kubewarden/rfc/blob/main/rfc/0016-wasm-cache.md).

This RFC explains how we can reduce the startup time of Policy Server by maintaining a cache of the policies and their optmized versions inside of a Kubernetes Persistent Volume.

### DIFF
--- a/rfc/0017-policy-server-reuse-wasm-optimized-modules.md
+++ b/rfc/0017-policy-server-reuse-wasm-optimized-modules.md
@@ -177,7 +177,7 @@ above.
 
 The init container would then run a program that performs the following steps:
 
-1. Obtain ownership of a lock via backed by a Kubernetes Lease resource.
+1. Obtain ownership of a lock backed by a Kubernetes Lease resource.
 2. Wait until exclusive ownership is obtained.
 3. Inspect the configuration of Policy Server, download all the missing `.wasm`
   modules
--- a/rfc/0017-policy-server-reuse-wasm-optimized-modules.md
+++ b/rfc/0017-policy-server-reuse-wasm-optimized-modules.md
@@ -201,9 +201,7 @@ The usage of an init container has the following benefits:
   Pod
 * Avoid further complexity to be added to the Policy Server codebase
 
-The "download & optimize" code would be provided by a dedicated binary that is
-executed by an init container.
-
+The "download & optimize" code would be provided by a dedicated binary.
 To keep things simple, this binary can be stored inside of the already existing
 Policy Server container image. By doing that we do not complicate our build pipeline
 and we do not introduce a new container image that has to be mirrored when
--- a/rfc/0017-policy-server-reuse-wasm-optimized-modules.md
+++ b/rfc/0017-policy-server-reuse-wasm-optimized-modules.md
@@ -1,0 +1,294 @@
+|              |                                                   |
+| :----------- | :------------------------------------------------ |
+| Feature Name | Policy Server: reuse optimized Wasm modules       |
+| Start Date   | Jul 18 2023                                       |
+| Category     | [Category]                                        |
+| RFC PR       | [PR22](https://github.com/kubewarden/rfc/pull/22) |
+| State        | **ACCEPTED**                                      |
+
+
+# Summary
+[summary]: #summary
+
+This RFC aims to provide a way to reduce the startup time of Policy Server.
+
+# Motivation
+[motivation]: #motivation
+
+Policy Server is restarted every time its configuration changes. This includes
+changes done to the policies that are deployed through it. This includes adding
+or removing policies, plus any kind of change done to an already deployed one
+(like a different configuration value, or a version bump).
+
+During the startup Policy Server performs these time expensive operations:
+
+* Download all the policies
+* Optimize the Wasm code for execution
+
+These operations are repeated every time, because the state of each Policy
+Server instance is not preserved across restarts.
+
+The purpose of this RFC is to find a way to reduce the startup time of
+Policy Server so that Kubewarden becomes more reactive to the changes done
+to the cluster policies.
+
+
+## Examples / User Stories
+[examples]: #examples
+
+> As a Kubernetes administrator using Kubewarden,
+> I want to see the new policy to be enforced
+> as soon as possible,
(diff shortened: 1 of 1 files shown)


rfc/0017-policy-server-reuse-wasm-optimized-modules.md
```diff
@@ -0,0 +1,294 @@
...
+files and all their optimized versions would be saved.
+
+The deployment would then define a new init container. The Policy Server
+configuration would be injected into the init container as a read-only file.
+The container would also have write access to the Persistent Volume described
+above.
+
+The init container would then run a program that performs the following steps:
+
+1. Obtain ownership of a lock via backed by a Kubernetes Lease resource.
```
viccuad:
```suggestion
1. Obtain ownership of a lock backed by a Kubernetes Lease resource.
```

rfc/0017-policy-server-reuse-wasm-optimized-modules.md
```diff
@@ -0,0 +1,294 @@
...
+almost immediately, given all the modules and their optimized versions are
+already inside of the Persistent Volume.
+
+The usage of an init container has the following benefits:
+
+* Create a clear representation of the bootstrap process of the Policy Server
+  Pod
+* Avoid further complexity to be added to the Policy Server codebase
+
+The "download & optimize" code would be provided by a dedicated binary that is
```
viccuad: I suppose this phrase is a bit redundant.

rfc/0017-policy-server-reuse-wasm-optimized-modules.md
```diff
@@ -0,0 +1,294 @@
...
+Development/testing environments should not be affected by that because
+k3s, minikube, kind and other similar projects provide a default Storage Class
+that is ready to be used.
+
+The majority of production environments already feature a Storage Class that
+can be reused.
+
+We can also change the helm chart to handle these possible configurations:
+
+* Use the default Storage Class
```
viccuad, line 228: We should be careful with this, as we can't ensure its write permissions I suppose.

rfc/0017-policy-server-reuse-wasm-optimized-modules.md
```diff
@@ -0,0 +1,294 @@
...
+We can reduce the impact of this dependency by creating a dedicated program that
+performs the download and optimization task. If `kubert` is abandoned we
+can write a Go program that takes the lock and then `exec` the "download & optimize"
+Rust code (which would no longer handle the lock aspects).
+
+## Garbage collector
+
+Not really a drawback, but something to keep in mind. Over the time the size of
+the Persistent Volume will grow. We will have to introduce a way to ensure old
+modules (and their optimized counterparts) are removed when no longer needed.
```
jvanz, line 250: Maybe, we can define a cronjob for that. Using the same binary used to optimize. But with different command line flag to clean the PV.

jvanz (approved):
It looks like fine to me. I believe we should do a POC and test this idea.

rfc/0017-policy-server-reuse-wasm-optimized-modules.md
flavio: What do you mean with that? I'll assume you are referring to being able to write into the PV.

The PV created by the StorageClass will be writable by any Pod. We can mount it as R/W inside of the init container, and then leave it as RO inside of the actual container running policy-server

rfc/0017-policy-server-reuse-wasm-optimized-modules.md
flavio: I would prefer to do that inside of the init container, it would prevent race conditions. The downloader can do a cleanup of the policies (and their optimized versions) that are not referenced by the current configuration AND are older than 1 week (a simple check on the creation timestamp)

flavio:
> It looks like fine to me. I believe we should do a POC and test this idea.

I already have a POC of the locking mechanism. I can extend that a little and try out the init container approach. Is that what you would like to see?

rfc/0017-policy-server-reuse-wasm-optimized-modules.md
flavio: Fixed, thanks

rfc/0017-policy-server-reuse-wasm-optimized-modules.md
viccuad: I was thinking that maybe other users could write to the PV. But PVs can only be bound to 1 PVC, and PVCs are namespaced. I suppose that one can only write to those PVs if one has admin rights (to mount and write directly in the node for example) or runs a privileged container (for which we already have the default policies).

Feel free to resolve the thread!

viccuad (approved):

jvanz:
> > It looks like fine to me. I believe we should do a POC and test this idea.
> 
> I already have a POC of the locking mechanism. I can extend that a little and try out the init container approach. Is that what you would like to see?

Never mind, this is enough indeed. I forgot that you are testing for a while already.

jvanz:
As far as I can see, we've reached an consensus. I'm merging this PR. Let me know if I've missed something.

rfc/0017-policy-server-reuse-wasm-optimized-modules.md
```diff
@@ -0,0 +1,292 @@
...
+## New deployment model
+
+Policy Server would continue to be deployed using a Deployment object.
+A new Persistent Volume would then be defined. This is where all the `.wasm`
+files and all their optimized versions would be saved.
+
+The deployment would then define a new init container. The Policy Server
+configuration would be injected into the init container as a read-only file.
+The container would also have write access to the Persistent Volume described
+above.
```
fabriziosestito, line 176: In a multi-node scenario, this only works if the storage class supports the `ReadWriteMany`. Otherwise, the volume is not shared between nodes.
Being `ReadWriteMany` is only supported by specific filesystems, the user should be aware of this.
Another approach could be having one volume per pod which will let us eliminate the lease. We could suggest to the user to isolate slow startup policies to a different policy server.

PR #24, opened 2023-08-11
RFC about certificate handling

flavio:
This RFC describes the current state of certificate handling inside of Kubewarden and how this should be extended.


rfc/0018-certificate-handling.md
```diff
@@ -0,0 +1,259 @@
+|              |                                 |
+| :----------- | :------------------------------ |
+| Feature Name | Certificate Handling            |
+| Start Date   | Aug 11 2023                     |
+| Category     | [Category]                      |
+| RFC PR       | [fill this in after opening PR] |
```
viccuad:
```suggestion
| RFC PR       | [#24](https://github.com/kubewarden/rfc/pull/24) |
```

rfc/0018-certificate-handling.md
```diff
@@ -0,0 +1,259 @@
+|              |                                 |
+| :----------- | :------------------------------ |
+| Feature Name | Certificate Handling            |
+| Start Date   | Aug 11 2023                     |
+| Category     | [Category]                      |
```
viccuad:
```suggestion
| Category     | enhancement, feature                      |
```

rfc/0018-certificate-handling.md
```diff
@@ -0,0 +1,259 @@
...
+by which components, which CA signed them and how have they been generated?
+
+The second goal is to come up with an architecture that removes the external
+dependency against CertManager. The new solution should also take into account
+certificate rotation.
+
+## Examples / User Stories
+
+> As a Kubernetes operator, I don't want to install CertManager
+> in order to deploy Kubearden.
```
viccuad:
```suggestion
> in order to deploy Kubewarden.
```

rfc/0018-certificate-handling.md
```diff
@@ -0,0 +1,259 @@
...
+
+- kubewarden-controller: this is a Validating and Mutating webhook endpoint that
+  processes the Kubewarden CRDs (like `PolicyServer`, `ClusterAdmissionPolicy`
+  and `AdmissionPolicy`)
+- Policy Server: each Deployment has its own dedicated certificate
+
+> **Note:** all these endpoints are either Validating or Mutating webhook
+> endpoints for Kubernetes
+
+Currently, Kubewarden 1.6 at the time of wriging, the certificate used by the
```
viccuad:
```suggestion
Currently, Kubewarden 1.6 at the time of writing, the certificate used by the
```

rfc/0018-certificate-handling.md
```diff
@@ -0,0 +1,259 @@
...
+- For each `PolicyServer` defined:
+  - Generate a new certificate, sign it with the new CA
+  - Force a rollout of the PolicyServer Deployment
+- Kubewarden Controller:
+  - Generate a new certificate, sign it with the new CA
+  - Restart the controller to ensure its HTTPS endpoint uses the new certificate
+- For each webhook configuration managed by Kubewarden (meaning all the policies deployed, plus the kubewarden-controller):
+  - Update the `clientConfig.caBundle`: ensure it contains the certificate of the new CA
+
+This change is massive and could lead to connection errors for as long as the PolicyServer are using a certificate
```
viccuad: Had the thought of configuring 2 caBundle on the WebhookClientConfig to perform the migration, but it only accepts 1 caBundle.

Also I wonder if we need an ordered approach here, first rotating the kubewarden-controller cert with the new CA, then, once the controller detects that its cert comes from the new CA, the reconciliation loops for webhook confs and PolicyServers. But I'm ok punting on this. Edit: read the unresolved questions section, I concur.

:+1:

viccuad (approved):
Ratified, just some small typo suggestions.

rfc/0018-certificate-handling.md
```diff
@@ -0,0 +1,259 @@
...
+Each `ClusterAdmissionPolicy` and `AdmissionPolicy` has a dedicated `(Validating|Mutating)WebhookConfiguration`
+object defined. For all of them, the `clientConfig.caBundle` contains the certificate of
+the CA generated by our controller.
+
+## Proposed Design
+
+We would like to get rid of CertManager. To do that we need to change how certificates are
+managed for the `kubewarden-controller`.
+Going forward, the controller will also take care of generating the certificate used by the kubewarden-controller.
+This certificate is going to be signed by the CA which is already created by the controller.
```
jvanz, line 97: Will we remove it completely or make it optional? I would like to double check that because the original issue around this topic says to turn it [optional](https://github.com/kubewarden/kubewarden-controller/issues/422#issuecomment-1644206549)

rfc/0018-certificate-handling.md
```diff
@@ -0,0 +1,259 @@
...
+- `PolicyServer`
+- `ClusterAdmissionPolicy`
+- `AdmissionPolicy`
+
+### Policy Server
+
+Reconciliation loops are triggered whenever an event dealing with a watched resource happens.
+In this case it could be the create/update/delete of a `PolicyServer`.
+The reconciliation loop is also triggered every X seconds as a way to cope with possible glitches
+with the event notification system.
```
jvanz, line 205: Where is this time based trigger setup? Is this done by kube-builder? I cannot see this  in the code. =(

jvanz (approved):
LGTM. I just have quick questions... ;)

rfc/0018-certificate-handling.md
viccuad: I agree on having it optional, and disabled by default.

fabriziosestito: +1

rfc/0018-certificate-handling.md
flavio: Good question, we're currently using [`controller-runtime`](https://github.com/kubernetes-sigs/controller-runtime) version 0.14.6.

When creating the Manager we give to it some configuration Options, see [here](https://github.com/kubewarden/kubewarden-controller/blob/fc6e1e815f87a80cedb737ca625604f5e80557d6/main.go#L135)

One of the values of the Manager Options is called [`SyncPeriod`](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.6/pkg/manager#Options). Since we don't specify anything, we will have an automatic sync interval of 10 hours. Meaning, even if there's no event affecting one of our watched resources, we will still have a sync every 10 hours.

This configuration parameter has been moved to another location inside of the latest release of controller-runtime. Now this is one of the [configuration values of the cache](https://github.com/kubernetes-sigs/controller-runtime/blob/d5bc8734caccabddac6a1bea250b0b9d771d318d/pkg/cache/cache.go#L160). The behaviour stays the same: every 10 hours you get a full sync.

We can rely on this 10 hours sync or we could be more explicit inside of our reconciliation loops by returning a  `reconcile.Result{RequeueAfter: t}` instead of a `reconcile.Result{}`.

What do you think?

fabriziosestito: I'd probably go with the latter.  As the docstring states, we shouldn't change the SyncPeriod as it forces every object in the cache to be reconciled, and we might need a shorter window than 10 hours. Not a strong opinion, though.

rfc/0018-certificate-handling.md
flavio: What would be the advantage of having this certificate generated by CertManager? I mean, we're still generating all the internal ones by ourselves.

I would prefer to drop the CertManager dependency entirely, just to reduce the support matrix and keep the code as simple as possible

rfc/0018-certificate-handling.md
flavio: I would either leave things untouched, hence rely on the default sync value, or provide an explicit sync timer via the `Result` response

rfc/0018-certificate-handling.md
```diff
@@ -0,0 +1,259 @@
...
+
+[unresolved]: #unresolved-questions
+
+The renewal of the root CA is the most disruptive event that can happen. Letting the reconciler loops run and wait
+for the system to converge towards the desired state is not a viable option. During the convergence
+time there could be requests being rejected due to Policy Server instances not being available or be using
+a certificate signed by an unexpected root CA.
+
+We must do some research and find a way to reduce this disruption. However, given our root CA has an expiration
+time of 10 years, this is a problem that should not block the implementation of this RFC.
```
flavio: I did some research, it's possible to renew a CA without changing its private key. In this way all the certificates issued by it are still going to be valid.
However, this is sometimes considered not a good security practice, it's recommended to create a new private key to: 1) push forward the cracking attempts of an attacker 2) eventually upgrade to a bigger key, better algorithm.

There's [this page](https://docs.aws.amazon.com/privateca/latest/userguide/ca-lifecycle.html) on AWS docs (search for "Reissuing an old CA") that provides this explanation, but I found similar opinions over the internet.

In our context, I think we should not worry too much about this scenario...

However, an easier solution to this problem could be the following one:

* create a new root CA
* create a CA bundle that contains the old CA and the new one
* update all the validation/mutation webook configurations to use the new CA bundle
* generate new certificates for all the policy servers, this time using the new root CA
* rollout the new certificates to the individual policy servers
* wait for the rollout to be complete
* create a new CA bundle that has only the new root CA
* update all the validation/mutation webook configurations to use the new CA bundle

I think this should work. What do you think? Let's discuss over there, I'll then incorporate the final outcome into the RFC

rfc/0018-certificate-handling.md
fabriziosestito: > I'd probably go with the latter. 

Sorry I did not explain myself well, I meant that I have a small preference for the Result solution.

rfc/0018-certificate-handling.md
fabriziosestito: I think this is the way to go. I found a similar approach suggested elsewhere (see for instance: https://www.cockroachlabs.com/docs/stable/rotate-certificates#rotate-the-ca-certificate)

rfc/0018-certificate-handling.md
flavio: No worries, that was clear to me. I also share the same feeling about how to move forward.

This change (being explicit about the sync time) should be tracked with a dedicated issue, which would be unrelated to this RFC

rfc/0018-certificate-handling.md
jvanz: I agree. This is a better approach. Let's do that.

flavio: Done with https://github.com/kubewarden/rfc/pull/24/commits/de8307cf3abad09d0a65a9674b2d301775b62d53

rfc/0018-certificate-handling.md
viccuad: @flavio does this comment still stand? I thought we wanted to still support cert-manager, but not as default.

viccuad (approved):

rfc/0018-certificate-handling.md
flavio: Last week @jvanz, @fabriziosestito and agreed to focus on removing the cert-manager dependency. We can make it optional in the future if there's some request coming from the community.

PR #26, opened 2023-09-01
docs: add policy fetcher database RFC

fabriziosestito:
This RFC describes a potential improvement to the policy-fetcher store, by using an embedded database.


flavio (commented):
It's good, but I think we should add a section under the "detailed design" explaining how the migration from the current storage system to the new one should be done

flavio (approved):
LGTM, thanks for the new section

jvanz (approved):

PR #33, opened 2024-03-26
docs: add reflectors database cache RFC

fabriziosestito:
## Description
As the result of https://github.com/kubewarden/policy-server/issues/716 spike, this PR adds an RFC proposing a disk-based database cache for the reflectors.

### DIFF
--- a/rfc/0019-reflectors-database-cache.md
+++ b/rfc/0019-reflectors-database-cache.md
@@ -66,6 +66,10 @@ The database will contain a table for each resource kind that is being watched b
 The table name will be in the following format: 'group_version_kind'.
 For example, the table name for the `Pod` resource kind will be `v1_pods` and the table name for the `Deployment` resource kind will be `apps_v1_deployments`.
 
+As SQLite does not support concurrent writes, the database will be accessed by a single writer thread and multiple reader threads.
+This could change in the future once `BEGIN CONCURRENT` will be supported.
+For the time being, if we hit concurrency issues, we could consider using a single database per resource type.
+
 Table schema:
 
 | Column    | Data Type    | Constraints |
--- a/rfc/0019-reflectors-database-cache.md
+++ b/rfc/0019-reflectors-database-cache.md
@@ -1,0 +1,150 @@
+|              |                                 |
+| :----------- | :------------------------------ |
+| Feature Name | Reflectors database cache       |
+| Start Date   | Mar 26 2024                     |
+| Category     | enhancement,feature             |
+| RFC PR       | [fill this in after opening PR] |
+| State        | **ACCEPTED**                    |
+
+# Summary
+
+[summary]: #summary
+
+This RFC proposes to offload the reflectors cache to a disk-based database to free up memory in the policy server.
+
+# Motivation
+
+[motivation]: #motivation
+
+Reflectors are a key component of the policy evaluator context-aware capabilities.
+They are used to keep track of the state of the Kubernetes cluster and to cache objects fetched from the Kubernetes API server that are needed to evaluate those policies that require information about the cluster state.
+
+The reflectors cache is currently stored in memory, which can lead to high memory usage in the policy server.
+
+The purpose of this RFC is to introduce a database into the reflectors to offload the cache to disk.
+Also, by using a database, we can implement queries that will reduce the number of reflectors that are created and maintained in memory.
+
+## Examples / User Stories
+
+[examples]: #examples
+
+- As a user, I want to reduce the memory usage of the policy server when caching objects from the Kubernetes API server.
+
+# Detailed design
+
+[design]: #detailed-design
+
+A [reflector](https://docs.rs/kube/latest/kube/runtime/reflector/fn.reflector.html) is a component that watches for changes in the Kubernetes API server and caches the objects that are relevant to the policy evaluation.
+It stores the watched objects in memory, and it is responsible for keeping them up-to-date.
+It also offers a reader interface to query the objects stored in the cache.
+At the time of writing, the reflectors are used to speed up the evaluation of policies that require information about the Kubernetes cluster state,
+by avoiding fetching the same objects from the Kubernetes API server multiple times during the evaluation of a single request.
+However, the reflectors cache is currently stored in memory, which can lead to high memory usage in the policy server.
+According to the [kube-rs ](https://kube.rs/controllers/optimization/#reflector-optimization)optimization[ guide](https://kube.rs/controllers/optimization/#reflector-optimization) and the [reflector documentation](https://docs.rs/kube/latest/kube/runtime/reflector/fn.reflector.html#memory-usage), the cache [Store](https://docs.rs/kube-runtime/0.88.1/src/kube_runtime/reflector/store.rs.html#13) is the main source of memory usage and it is directly proportional to how many objects are stored.
+In the worst-case scenario (Pods with many sidecar containers and many environment variables), the reflectors cache can consume up to 1GB of memory for ~2000 objects.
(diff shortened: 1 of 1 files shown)


rfc/0019-reflectors-database-cache.md
```diff
@@ -0,0 +1,164 @@
...
+[motivation]: #motivation
+
+Reflectors are a key component of the policy evaluator context-aware capabilities.
+They are used to keep track of the state of the Kubernetes cluster,
+and to cache objects fetched from the Kubernetes API server that are needed to evaluate those policies that require information about the cluster state.
+
+The reflectors cache is currently stored in memory, which can lead to high memory usage in the policy server.
+
+The purpose of this RFC is to introduce a database into the reflectors to offload the cache to disk.
+Also, by using a database, we can implement queries that will reduce the number of reflectors that are created and maintained in memory.
```
flavio, line 25: Can you also add that reflectors rely on kubernetes watches, which are an expensive resource. The more watches are created against an API server, the more load you put on it

rfc/0019-reflectors-database-cache.md
```diff
@@ -0,0 +1,164 @@
...
+
+[examples]: #examples
+
+<!---
+Examples of how the feature will be used. Interactions should show the action
+and the response. When appropriate, provide user stories in the form of "As a
+[role], I want [feature], so [that]."
+--->
+
+See other RFCs.
```
flavio: ?

rfc/0019-reflectors-database-cache.md
```diff
@@ -0,0 +1,164 @@
...
+It stores the watched objects in memory, and it is responsible for keeping them up-to-date.
+It also offers a reader interface to query the objects stored in the cache.
+At the time of writing, the reflectors are used to speed up the evaluation of policies that require information about the Kubernetes cluster state,
+by avoiding fetching the same objects from the Kubernetes API server multiple times during the evaluation of a single request.
+However, the reflectors cache is currently stored in memory, which can lead to high memory usage in the policy server.
+According to the [kube-rs optimizian guide](https://kube.rs/controllers/optimization/#reflector-optimization) and the [reflector documentation](https://docs.rs/kube/latest/kube/runtime/reflector/fn.reflector.html#memory-usage), the cache [Store](https://docs.rs/kube-runtime/0.88.1/src/kube_runtime/reflector/store.rs.html#13) is the main source of memory usage and it is directly proportional to how many objects are stored.
```
flavio:
```suggestion
According to the [kube-rs optimization guide](https://kube.rs/controllers/optimization/#reflector-optimization) and the [reflector documentation](https://docs.rs/kube/latest/kube/runtime/reflector/fn.reflector.html#memory-usage), the cache [Store](https://docs.rs/kube-runtime/0.88.1/src/kube_runtime/reflector/store.rs.html#13) is the main source of memory usage and it is directly proportional to how many objects are stored.
```

rfc/0019-reflectors-database-cache.md
```diff
@@ -0,0 +1,164 @@
...
+
+## SQLite database store
+
+This RFC proposes to create a custom `Store` that uses a disk-based database to store the objects.
+The database will be used to store the objects in JSON format, and it will be queried using the `fieldSelector` and `labelSelector` to reduce the number of reflectors that are created and maintained in memory.
+SQLite is a good candidate for the database, as it is a single file database and supports SQL and JSON queries.
+Also, the database can be considered ephemeral, as the reflectors can be recreated from the Kubernetes API server if needed.
+The database will contain a table for each resource kind that is being watched by the reflectors.
+
+Table schema:
```
flavio, line 73: You need to mention we're going to create one table per type of watched request. Plus explain the naming convention we want to use to name these tables.

Finally, point out that by having 1 table per resource we reduce the concurrency issues when inserting/querying data.

flavio (changes requested):
LGTM, I left some comments that need to be addressed

rfc/0019-reflectors-database-cache.md
```diff
@@ -0,0 +1,164 @@
...
+The inventory is built at boot time and kept up-to-date by watching for changes in the Kubernetes API server.
+
+# Drawbacks
+
+[drawbacks]: #drawbacks
+
+At boot time the policy server will start the reflectors.
+Once a reflector is started it will receive a `watcher::Event::Restarted` event with all the objects of the resource kind it is watching.
+Even if the request is paginated (see: https://docs.rs/kube/latest/kube/runtime/watcher/struct.Config.html#method.page_size)
+to reduce Kubernetes API and serialization load, the items will be buffered-up in the watcher event causing a spike in memory usage.
```
jvanz: As far as I can see, we will not wait for the initial load of the database. Therefore, I think that another drawback is that right after the starting of the policy server, if it receives a evaluation request that requires many resources that are still being loaded,  the policy can evaluate using incomplete data. Because the reflectors will be still loading the data into the database. Leading to invalid resources being accepted in the cluster. 

I don't believe that will be a huge problem now. Because we also have the audit scanner to inspect the cluster. However, I believe this can be important for users that have policies that requires access to massive resources count. Thus,  I think is important to document that.

jvanz (changes requested):

rfc/0019-reflectors-database-cache.md
```diff
@@ -0,0 +1,164 @@
...
+Get all the pods in the `default` namespace:
+
+```sql
+SELECT object FROM v1_pods WHERE namespace = 'default';
+```
+
+Get all the pods with the label `app=nginx`:
+
+```sql
+SELECT object from v1_pods where json_extract(object, '$.metadata.labels.app') = 'nginx';
```
jhkrug: Just for consistency.
```suggestion
SELECT object FROM v1_pods WHERE json_extract(object, '$.metadata.labels.app') = 'nginx';
```

rfc/0019-reflectors-database-cache.md
```diff
@@ -0,0 +1,164 @@
...
+```
+
+## Gatekeeper inventory cache
+
+Gatekeeper policies require a different approach, as they expect the cache data to be present under the `data.inventory` document.
+To know more about the Gatekeeper inventory document, please refer to the [Gatekeeper documentation](https://open-policy-agent.github.io/gatekeeper/website/docs/sync/#accessing-replicated-data).
```
jhkrug: I think it reads better.
```suggestion
To learn more about the Gatekeeper inventory document, please refer to the [Gatekeeper documentation](https://open-policy-agent.github.io/gatekeeper/website/docs/sync/#accessing-replicated-data).
```

jhkrug (approved):
Two tiny things, but lgtm.

flavio (approved):

rfc/0019-reflectors-database-cache.md
fabriziosestito: @jvanz clarified in https://github.com/kubewarden/rfc/pull/33/commits/4195fb6edacbe7f6693015143ee52977026c71aa

viccuad (approved):

rfc/0019-reflectors-database-cache.md
```diff
@@ -0,0 +1,154 @@
+|              |                                 |
+| :----------- | :------------------------------ |
+| Feature Name | Reflectors database cache       |
+| Start Date   | Mar 26 2024                     |
+| Category     | enhancement,feature             |
+| RFC PR       | [fill this in after opening PR] |
```
flavio: Please update this field and then merge the PR

flavio (approved):

jvanz:
:tada:

PR #35, opened 2024-05-20
docs: reject RFC 0019

fabriziosestito:
## Description

This PR rejects the RFC 0019 (Reflectors database cache).
It adds the motivation of the rejection and the benchmark results.


flavio (approved):
That was quite a ride, thanks for the hard work.

It's a bit disappointing to see there's actually no silver bullet :)

jvanz (approved):

PR #37, opened 2024-06-13
docs(rfc): add policy group RFC

### LINKED ISSUE #34: RFC: how to group multiple policies using logical operators
We would like to combine multiple policies together using logical operators.

For example: it would `pod is not running as root AND NOT image registry == "infra.registry.local.lan"`

fabriziosestito:
## Description

Adds policy group RFC

Fixes #34

### DIFF
--- a/rfc/0020-group-policies.md
+++ b/rfc/0020-group-policies.md
@@ -1,0 +1,263 @@
+|              |                                 |
+| :----------- | :------------------------------ |
+| Feature Name | Group policies                  |
+| Start Date   | Jun 12 2024                     |
+| Category     | feature                         |
+| RFC PR       | [fill this in after opening PR] |
+| State        | **ACCEPTED**                    |
+
+# Summary
+
+A "group policy" is a policy that is composed of other policies.
+The policies that are part of a group are evaluated using a boolean expression.
+The group policy will not support mutations.
+
+# Motivation
+
+[motivation]: #motivation
+
+The motivation for this feature is to enable users to create complex policies by combining simpler ones. This allows users to avoid the need to create custom policies from scratch and instead leverage existing policies.
+This reduces the need to duplicate policy logic across different policies, increases reusability, removes the cognitive load of managing complex policy logic,
+and enables the creation of custom policies using a DSL-like configuration.
+
+## Examples / User Stories
+
+[examples]: #examples
+
+- As a user, I want to create a policy that is composed of other policies and is evaluated using a boolean expression.
+
+# Detailed design
+
+[design]: #detailed-design
+
+## Controller and CRDs
+
+This RFC proposes to add two new CRDs to the Kubewarden controller:
+
+- `GroupAdmissionPolicy`
+- `GroupClusterAdmissionPolicy`
+
+The CRDs will share most of the fields with the `AdmissionPolicy` and `ClusterAdmissionPolicy` CRDs, respectively.
+The main difference is that the `GroupAdmissionPolicy` and `GroupClusterAdmissionPolicy` CRDs will have a list of policies.
+Also, the `GroupAdmissionPolicy` and `GroupClusterAdmissionPolicy` CRDs will have a field to specify the boolean expression that will be used to evaluate the policies
+and a `message` field to specify the message that will be returned when the group policy is rejected.
+
+## Example CRD
+
+```yaml
+apiVersion: policies.kubewarden.io/v1
+kind: ClusterAdmissionPolicyGroup # or AdmissionPolicyGroup
+metadata:
+  name: pod-image-signatures
+spec:
+  rules:
+    - apiGroups: [""]
+      apiVersions: ["v1"]
+      resources: ["pods"]
+      operations:
+        - CREATE
+        - UPDATE
+  policies:
+    - name: sigstore_pgp
+      module: ghcr.io/kubewarden/policies/verify-image-signatures:v0.2.8
+      settings:
+        signatures:
+          - image: "*"
+            pubKeys:
+              - "-----BEGIN PUBLIC KEY-----xxxxx-----END PUBLIC KEY-----"
+              - "-----BEGIN PUBLIC KEY-----xxxxx-----END PUBLIC KEY-----"
+    - name: sigstore_gh_action
+      module: ghcr.io/kubewarden/policies/verify-image-signatures:v0.2.8
+      settings:
+        signatures:
+          - image: "*"
+            githubActions:
+            owner: "kubewarden"
+    - name: reject_latest_tag
+      module: ghcr.io/kubewarden/policies/trusted-repos-policy:v0.1.12
+      settings:
+        tags:
+          reject:
+            - latest
+  expression: "sigstore_pgp() || (sigstore_gh_action() && reject_latest_tag())"
+  message: "The group policy is rejected."
+```
+
+### Expression language
+
+We will use [CEL](https://github.com/google/cel-go) as the expression language for the group policies.
+The main reason for this choice is that CEL is used by the [ValidatingAdmissionPolicy](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/) and [matchConditions](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-matchconditions) in Kubernetes.
+
+Each policy in the group will be represented as a function call in the expression with the same name as the policy defined in the group.
+The expression field should be a valid CEL expression that evaluates to a boolean value and it will be validated by the Kubewarden controller's webhook.
+If the expression evaluates to `true`, the group policy will be considered as `accepted`, otherwise, it will be considered as `rejected`.
+
+### Message
+
+The message field will be used to specify the message that will be returned when the group policy is rejected.
+The specific policy results will be returned in the `warning` field of the response.
+
+This is an example of the response that will be returned when the group policy is rejected:
+
+```json
+{
+  "apiVersion": "admission.k8s.io/v1",
+  "kind": "AdmissionReview",
+  "response": {
+    "uid": "<value from request.uid>",
+    "allowed": false,
+    "warning": [
+      "sigstore_pgp was rejected: the image signature is not valid",
+      "sigstore_gh_action was accepted",
+      "latest_tag was rejected: the image tag latest is not allowed"
+    ]
+  }
+}
+```
+
+### Reconciliation
+
+The new CRDs will be reconciled by the Kubewarden controller.
+The reconciliation flow is similar to the one used for the `AdmissionPolicy` and `ClusterAdmissionPolicy` resources:
+
+- when a new `GroupAdmissionPolicy` or `GroupClusterAdmissionPolicy` is created or updated the reconciler changes the configuration of the policy server to include the created/updated policy group
+- the reconciler rolls out the policy server deployment
+- the reconciler creates or updates the ValidatingWebhookConfiguration pointing to the policy server `validate` endpoint of the policy group
+
+### Match Conditions
+
+An interesting use case is to use the group policies in combination with the [matchConditions](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-matchconditions) feature.
+
+By combining the group policies with `matchConditions`, it could be possible to create complex policies that are evaluated only for specific requests.
+For instance, it could be possible to build a policy that is evaluated only for requests that match a specific label selector.
+
+This RFC does not include any modifications to the policy CRDs to accommodate `matchConditions`, as it falls outside the scope of this proposal.
+
+## Policy server
+
+The policy settings will be extended to include group policies alongside single policies.
+
+```yaml
+psp-apparmor: # single policy
+  url: registry://ghcr.io/kubewarden/policies/psp-apparmor:v0.1.7
+
+psp-capabilities: # single policy
+  url: registry://ghcr.io/kubewarden/policies/psp-capabilities:v0.1.7
+  allowedToMutate: true
+  settings:
+    allowed_capabilities: ["*"]
+    required_drop_capabilities: ["KILL"]
+
+pod-image-singatures: # group policy
+  policies:
+    - name: sigstore_pgp
+      module: ghcr.io/kubewarden/policies/verify-image-signatures:v0.2.8
+      settings:
+        signatures:
+          - image: "*"
+            pubKeys:
+              - "-----BEGIN PUBLIC KEY-----xxxxx-----END PUBLIC KEY-----"
+              - "-----BEGIN PUBLIC KEY-----xxxxx-----END PUBLIC KEY-----"
+    - name: sigstore_gh_action
+      module: ghcr.io/kubewarden/policies/verify-image-signatures:v0.2.8
+      settings:
+        signatures:
+          - image: "*"
+            githubActions:
+            owner: "kubewarden"
+    - name: reject_latest_tag
+      module: ghcr.io/kubewarden/policies/trusted-repos-policy:v0.1.12
+      settings:
+        tags:
+          reject:
+            - latest
+  expression: "sigstore_pgp() || (sigstore_gh_action() && reject_latest_tag())"
+  message: "The group policy is rejected."
+```
+
+### Expression evaluation
+
+Unfortunately, no production-ready CEL library is available for Rust.
+To emulate the behavior and the syntax of CEL, we will use [Rhai](https://rhai.rs/), a small, fast, easy-to-use scripting language and evaluation engine that integrates tightly with Rust.
+
+Rhai has the following features that make it a good candidate for this task:
+
+- function calls have the same syntax as CEL
+- it is easy to embed in a Rust application
+- it provides a way to create a [minimal engine](https://rhai.rs/book/engine/raw.html) with a bare minimum set of functionalities
+- it does not add a considerable overhead (see [Preliminary benchmark](#preliminary-benchmark))
+- the language supports [syntax customization](https://rhai.rs/book/engine/custom-syntax.html) if needed
+
+Considering the following expression from the example above:
+
+```rust
+sigstore_pgp() || (sigstore_gh_action() && reject_latest_tag())
+```
+
+Each policy in the group will be represented as a function call in the expression with the same name as the policy defined in the group.
+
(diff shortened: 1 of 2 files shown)


rfc/0020-group-policies.md
```diff
@@ -0,0 +1,263 @@
...
+[drawbacks]: #drawbacks
+
+- The implementation of the group policies will add complexity to the Kubewarden controller and the policy server.
+- The expression language used to evaluate the group policies may not be familiar to all users.
+- In the worst case, the group policy evaluation time could be the sum of the evaluation time of all the policies in the group.
+  This is even worse if the group policy contains context-aware policies that introduce additional overhead.
+
+# Alternatives
+
+[alternatives]: #alternatives
```
fabriziosestito: not sure we even considered an alternative.
should I mention the wasm bundle idea?

rfc/0020-group-policies.md
```diff
@@ -0,0 +1,263 @@
...
+An interesting use case is to use the group policies in combination with the [matchConditions](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-matchconditions) feature.
+
+By combining the group policies with `matchConditions`, it could be possible to create complex policies that are evaluated only for specific requests.
+For instance, it could be possible to build a policy that is evaluated only for requests that match a specific label selector.
+
+This RFC does not include any modifications to the policy CRDs to accommodate `matchConditions`, as it falls outside the scope of this proposal.
+
+## Policy server
+
+The policy settings will be extended to include group policies alongside single policies.
```
fabriziosestito: I've used `single` policy all over the RFC, maybe there is a better name for it.

jvanz: Maybe "ordinary" policy or just "policy".

rfc/0020-group-policies.md
jvanz: I'm curious about that wasm bundle idea. I would love to now more about it.

jvanz (commented):
At the first glace, it looks nice to me. But I would like to learn more about Rhai before my final opinion. I'll study a little bit more about it.

rfc/0020-group-policies.md
```diff
@@ -0,0 +1,263 @@
...
+            pubKeys:
+              - "-----BEGIN PUBLIC KEY-----xxxxx-----END PUBLIC KEY-----"
+              - "-----BEGIN PUBLIC KEY-----xxxxx-----END PUBLIC KEY-----"
+    - name: sigstore_gh_action
+      module: ghcr.io/kubewarden/policies/verify-image-signatures:v0.2.8
+      settings:
+        signatures:
+          - image: "*"
+            githubActions:
+            owner: "kubewarden"
```
flavio: I think this is the proper indentation, I love YAML

```suggestion
              owner: "kubewarden"
```

rfc/0020-group-policies.md
flavio: I guess you're referring to the `policies.yaml` file used by the Policy Server? If that's correct, I would just put the name of the file in there.

rfc/0020-group-policies.md
```diff
@@ -0,0 +1,263 @@
...
+Note that the `sigstore_gh_action` function will be called only if the `sigstore_pgp` function returns `false`.
+Similarly, the `reject_latest_tag` function will be called only if the `sigstore_gh_action` function returns `true`.
+
+This avoids evaluating all the policies in the group when the result is already known.
+
+### Settings validation
+
+When the policy server starts, it will validate the settings of the group policies as well as the single policies.
+However, the group policies will have an additional validation step to ensure that the expression is valid and
+that evaluates to a boolean value.
```
flavio: I think we should also look for a misconfiguration like mixing policies that look into totally different resources. For example, there's no sense in doing `policy_that_eval_ingress() && policy_that_eval_pods()`.

In this case I think we should print a big fat warning

rfc/0020-group-policies.md
```diff
@@ -0,0 +1,263 @@
...
+- In the worst case, the group policy evaluation time could be the sum of the evaluation time of all the policies in the group.
+  This is even worse if the group policy contains context-aware policies that introduce additional overhead.
+
+# Alternatives
+
+[alternatives]: #alternatives
+
+# Unresolved questions
+
+[unresolved]: #unresolved-questions
```
flavio: Do we plan to do some validation when a `ClusterAdmissionPolicyGroup` is created?

Things like:

1. ensure the expression is valid
2. ensure the policies mixed together evaluate compatible types of resources

The 1st issue could be addressed, but there's the problem that Rhai is not available for Go. We could use CEL, given we plan to customize Rhai to look and behave like CEL, but there's no way to strip away CEL functions from it. What I mean is that `"foo".startsWith("f") && policy_1() || policy_2()` is a good CEL expression, but we would panic later on when evaluating that inside of Rhai.

Maybe a solution to this problem would be to create a Wasm module that contains our Rhai validator and use that inside of the controller. That would be quite cool :sunglasses: 


As for the 2nd issue. I fear there's nothing we can do on the short term. We might want to introduce a sidecar in the future that pulls the policies and inspects their metadata.

rfc/0020-group-policies.md
flavio: The idea was to create a tool that takes N-policies and merges them into a new one. The new Wasm module would embed the original ones and would have a "main" function that takes care of evaluating the expression given by the user.

The main drawbacks here hare:
* Complexity to implement this solution
* We would require an extra "compilation" step to be done by our users
* At runtime, the amount of memory saved would be significantly reduced. With this proposal, 2 group policies using the same building blocks would reuse the same `EvaluationContext` resources. With the "wasm bundle" this would not happen

flavio (approved):
LGTM, thanks for the detailed work

rfc/0020-group-policies.md
```diff
@@ -0,0 +1,263 @@
+|              |                                 |
+| :----------- | :------------------------------ |
+| Feature Name | Group policies                  |
+| Start Date   | Jun 12 2024                     |
+| Category     | feature                         |
+| RFC PR       | [fill this in after opening PR] |
```
viccuad:
```suggestion
| RFC PR       | https://github.com/kubewarden/rfc/pull/37 |
```

rfc/0020-group-policies.md
```diff
@@ -0,0 +1,263 @@
...
+The CRDs will share most of the fields with the `AdmissionPolicy` and `ClusterAdmissionPolicy` CRDs, respectively.
+The main difference is that the `GroupAdmissionPolicy` and `GroupClusterAdmissionPolicy` CRDs will have a list of policies.
+Also, the `GroupAdmissionPolicy` and `GroupClusterAdmissionPolicy` CRDs will have a field to specify the boolean expression that will be used to evaluate the policies
+and a `message` field to specify the message that will be returned when the group policy is rejected.
+
+## Example CRD
+
+```yaml
+apiVersion: policies.kubewarden.io/v1
+kind: ClusterAdmissionPolicyGroup # or AdmissionPolicyGroup
```
viccuad: This naming is inconsistent with the previous paragraph. I highly prefer this one though.
In any case, I would prefer one that starts with `Admission` and `ClusterAdmission` rather than `Group` and `GroupCluster`.

rfc/0020-group-policies.md
```diff
@@ -0,0 +1,263 @@
...
+The main reason for this choice is that CEL is used by the [ValidatingAdmissionPolicy](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/) and [matchConditions](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-matchconditions) in Kubernetes.
+
+Each policy in the group will be represented as a function call in the expression with the same name as the policy defined in the group.
+The expression field should be a valid CEL expression that evaluates to a boolean value and it will be validated by the Kubewarden controller's webhook.
+If the expression evaluates to `true`, the group policy will be considered as `accepted`, otherwise, it will be considered as `rejected`.
+
+### Message
+
+The message field will be used to specify the message that will be returned when the group policy is rejected.
+The specific policy results will be returned in the `warning` field of the response.
```
viccuad: Instead of warning messages, I would prefer to populate the admissionReview response.status, as it has  an array `statusDetails.causes` that seems fitting: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#statusdetails-v1-meta.

rfc/0020-group-policies.md
viccuad: I understand it as referring to usual (Cluster)AdmissionPolicies.
I would start naming things as follows: (Cluster)AdmissionPolicyGroups and (Cluster)AdmissionPolicies:

```suggestion
The policy settings will be extended to include PolicyGroups alongside single Policies.
```

rfc/0020-group-policies.md
```diff
@@ -0,0 +1,263 @@
...
+  url: registry://ghcr.io/kubewarden/policies/psp-apparmor:v0.1.7
+
+psp-capabilities: # single policy
+  url: registry://ghcr.io/kubewarden/policies/psp-capabilities:v0.1.7
+  allowedToMutate: true
+  settings:
+    allowed_capabilities: ["*"]
+    required_drop_capabilities: ["KILL"]
+
+pod-image-singatures: # group policy
```
viccuad:
```suggestion
pod-image-signatures: # group policy
```

rfc/0020-group-policies.md
viccuad: I would be glad to have the wasm bundled listed as alternative.

I suppose it's worth to note the future alternative of having a proper Rust CEL interpreter and ordering execution of policies by execution cost, to reduce the time on shortcircuit.

rfc/0020-group-policies.md
```diff
@@ -0,0 +1,263 @@
...
+- it does not add a considerable overhead (see [Preliminary benchmark](#preliminary-benchmark))
+- the language supports [syntax customization](https://rhai.rs/book/engine/custom-syntax.html) if needed
+
+Considering the following expression from the example above:
+
+```rust
+sigstore_pgp() || (sigstore_gh_action() && reject_latest_tag())
+```
+
+Each policy in the group will be represented as a function call in the expression with the same name as the policy defined in the group.
```
viccuad: Do we want in the future to swap Rhai for a CEL interpreter and add CEL `messageExpressions`?

rfc/0020-group-policies.md
```diff
@@ -0,0 +1,263 @@
...
+and a `message` field to specify the message that will be returned when the group policy is rejected.
+
+## Example CRD
+
+```yaml
+apiVersion: policies.kubewarden.io/v1
+kind: ClusterAdmissionPolicyGroup # or AdmissionPolicyGroup
+metadata:
+  name: pod-image-signatures
+spec:
```
viccuad: how do we define context awareness? From the implementation, it can only be on each policy inside of the `policies` array, but I think it's worth mentioning. This means that tools need to traverse the PolicyGroup spec to see if a policy is context aware.

viccuad: We haven't mentioned `backgroundAudit`, yet I understand PolicyGroups support it.

rfc/0020-group-policies.md
flavio: I also prefer `ClusterAdmissionPolicyGroup`

rfc/0020-group-policies.md
flavio: Good point, I think we should extend the RFC to cover also context aware policies. In this scenario, I think it would be easier (from a UX POV) to have just one place where this is defined

flavio: Yes, this should work out of the box. It's worth to be written inside of the RFC

rfc/0020-group-policies.md
jvanz: Interesting. But I prefer the current proposal as well.

rfc/0020-group-policies.md
jvanz: > I also prefer ClusterAdmissionPolicyGroup

Same here

rfc/0020-group-policies.md
jvanz: If we have a single place to define the resources, and considering that each policy will have its wasm module,  cannot we give access to policies that should not have access to resource?

rfc/0020-group-policies.md
fabriziosestito: we need to investigate if we can use a subset of CEL: https://github.com/google/cel-go/issues/899  to disable everything but boolean operators and policy functions.

rfc/0020-group-policies.md
fabriziosestito: @viccuad  I've changed to "policy group(s)" . I am avoiding the camel-case as we don't have a resource named `Policy` or `PolicyGroup` (without (Cluster)Admission) and this could be confusing.

rfc/0020-group-policies.md
fabriziosestito: unfortunately there is no production ready CEL rust interpreter so I am going to add this in the unresolved questions section

rfc/0020-policy-group.md
```diff
@@ -0,0 +1,284 @@
...
+
+[motivation]: #motivation
+
+The motivation for this feature is to enable users to create complex policies by combining simpler ones. This allows users to avoid the need to create custom policies from scratch and instead leverage existing policies.
+This reduces the need to duplicate policy logic across different policies, increases reusability, removes the cognitive load of managing complex policy logic,
+and enables the creation of custom policies using a DSL-like configuration.
+
+## Examples / User Stories
+
+[examples]: #examples
```
viccuad, line 25: Another example worth adding, with its ramifications:
> As a user, I want to iterate on developing a policy group locally, and test it using kwctl.

rfc/0020-group-policies.md
viccuad: I would prefer to return results using this approach.

viccuad: bump, just in case it was lost in the thread.

rfc/0020-policy-group.md
```diff
@@ -0,0 +1,284 @@
...
+The new WASM module would embed the original ones and would have a `main` function that takes care of evaluating the expression given by the user.
+
+This approach has the following drawbacks:
+
+- Complexity to implement this solution
+- The user would require an extra compilation step to create the new WASM module
+- The size of the new WASM module would be significantly larger if several policies are embedded
+- At runtime, the de-duplication optmization of the policies would be lost, increasing the memory footprint
+
+# Unresolved questions
```
viccuad, line 328: Are policy groups compatible with raw policies? (that return a json)
Do we foresee supporting policies in policy groups that return valid CEL, to be evaluated in `expression`?
It seems that's the case, and this implementation API is forward compatible for that, hence maybe worth a mention.

fabriziosestito: policy group are fully compatbile with raw policies out of the box, I will add a paragraph to clarify this

rfc/0020-group-policies.md
fabriziosestito: ~I think that a global definition is enough as the policies in the group are acting as a single policy from a security standpoint. I am leaving this conversation unresolved if there's the need to discuss further about this.~

We decided to have local policy context-aware rules instead.

rfc/0020-group-policies.md
fabriziosestito: @viccuad I haven't but we can have a discussion about this during the Parking lot today so I will leave this unresolved.

rfc/0020-policy-group.md
```diff
@@ -275,6 +276,11 @@ Also, we could consider creating a [custom package](https://rhai.rs/book/rust/pa
 The policy server will support [raw policy](https://docs.kubewarden.io/tutorials/writing-policies/wasi/raw-policies) groups out of the box, since the group evaluation logic is implemented in the `EvaluationEnvironment`.
 It will be possible to define a policy group with raw policies only, and evaluate the expression by calling the `validate_raw/<group name>` endpoint.
 
+## Development
+
+The `kwctl` tool will be extended to support the policy groups, to allow users to develop and test policy groups locally.
```
fabriziosestito: @kubewarden/kubewarden-developers 
it's worth to mention how the `kwctl` cli will run policy groups. 
Do you have any idea on how this could look like?

rfc/0020-policy-group.md
fabriziosestito: https://github.com/kubewarden/rfc/pull/37/commits/1ec324500388de36436842bc67e7f7aa8888e134#r1644306056

rfc/0020-policy-group.md
fabriziosestito: forgot to mention scaffolding

rfc/0020-group-policies.md
fabriziosestito: It appears that `statusDetails.causes` is designed to return errors. See usage here: https://github.com/googleforgames/agones/blob/541dafe5684db40a169cdffdd5a16a4023260c20/pkg/util/webhooks/webhooks.go#L103 . 
We want instead to provide a list of information about the policy evaluation result. 
As VAP is using the warning fields for the failed expression, I think we should go with this one. (see https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/#validation-actions)

viccuad: Good find. Indeed, I agree, let's go with the warning fields. Resolving.

rfc/0020-policy-group.md
```diff
@@ -0,0 +1,344 @@
...
+# Unresolved questions
+
+[unresolved]: #unresolved-questions
+
+- Unfortuantely, no production-ready CEL library is available for Rust.
+  Some experimentes were made with [cel-rust](https://github.com/clarkmcc/cel-rust) and [rscel](https://github.com/1BADragon/rscel)
+  but they do not pass the official compliance tests yet and are not actively developed.
+
+- Furthermore, using two different expression languages in the Kubewarden controller and the policy server could lead to inconsistencies in the validation step.
+  For instance, it is possibilie that an expression that is valid in the Kubewarden controller is not valid in the policy server.
```
viccuad:
```suggestion
  For instance, it is possible that an expression that is valid in the Kubewarden controller is not valid in the policy server.
```

viccuad (approved):
LGTM!

jvanz (approved):

PR #42, opened 2024-09-11
docs: update policy group rfc

fabriziosestito:
## Description

Updates the policy group RFC with the right CRD/config format of policy group members.


jhkrug (approved):
ok

PR #45, opened 2025-02-03
feat: Define policy Helm charts further

viccuad:
## Description

Relates to https://github.com/rancher/kubewarden-ui/issues/1024

This PR attempts to document all the necessary features of a policy Helm chart.

## Test

To test this pull request, you can run the following commands:

### DIFF
--- a/rfc/0010-rancher-integration-policies.md
+++ b/rfc/0010-rancher-integration-policies.md
@@ -1,13 +1,13 @@
-|              |                                  |
-| :----------- | :------------------------------- |
-| Feature Name | Rancher integration of Kubewarden Policies  |
-| Start Date   | 2022-08-22                       |
-| Category     | Development                      |
+|              |                                            |
+| :----------- | :----------------------------------------- |
+| Feature Name | Rancher integration of Kubewarden Policies |
+| Start Date   | 2022-08-22                                 |
+| Category     | Development                                |
 | RFC PR       | https://github.com/kubewarden/rfc/pull/11  |
-| State        | **ACCEPTED**                     |
-
+| State        | **ACCEPTED**                               |
 
 # Summary
+
 [summary]: #summary
 
 This RFC explains how the Kubewarden policies, either owned by the Kubewarden
@@ -20,8 +20,8 @@ For 3rd party policies not owned by the Kubewarden team,
 it proposes using the Artifact Hub API to poll for the metadata stored in
 `policy/artifacthub-pkg.yml`.
 
-
 # Motivation
+
 [motivation]: #motivation
 
 Problems to solve:
@@ -33,6 +33,7 @@ Problems to solve:
 - Allow for airgap installations.
 
 ## Examples / User Stories
+
 [examples]: #examples
 
 1. As a user, I want to install one Kubewarden policy owned by the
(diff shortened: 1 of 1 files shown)


rfc/0010-rancher-integration-policies.md
```diff
@@ -47,34 +48,92 @@ Problems to solve:
...
+### Policy distribution
 
-Even if the charts will not be released to users, the repository can be consumed
+Submit all Helm charts of the policies to a new repository,
+https://github.com/kubewarden/policy-charts, under a `charts/` folder. This
+repository will contain similar CI and scripts as the kubewarden/helm-charts
+repository. The resulting policy Helm charts will get shipped into a separate
+HTTPs Helm repository along with an OCI repository in ghcr.io.
+
+Even if the charts will not be publitized to users, the repository can be consumed
```
flavio:
```suggestion
Even if the charts will not be publicized to users, the repository can be consumed
```

flavio (approved):
LGTM

rfc/0010-rancher-integration-policies.md
```diff
@@ -47,34 +48,92 @@ Problems to solve:
...
+- `io.artifacthub.keywords`
+- `io.artifacthub.resources`
+
+It is valid to just duplicate all annotations present in the metadata.yml of
+the policy in question, for simplicity.
+
+The following annotations are created from the metadata.yml entries of the policy:
+
+- `kubewarden/mutating`
+- `kubewarden/contextAware`
```
jordojordo: I have updated my PR of the POC, I've made a few changes to the annotations I'm using within the UI:

We can replicate what we were doing in the metadata.yml for the policies with a few changes, namely the description and displayName. Currently I'm using the `description` property from the Chart.yaml in place of the `io.kubewarden.policy.description`. Also I'm open to changing the annotation names, this is just what I was using for testing:

```yaml
  # Annotations needed for the UI
  kubewarden/displayName: <policy-displayName>
  kubewarden/mutation: 'true|false'
  kubewarden/contextAwareResources: 'true|false'
  kubewarden/resources: <comma-separated-string>
  ```

File: rfc/0010-rancher-integration-policies.md [hunk 3]
```diff
@@ -47,34 +48,92 @@ Problems to solve:
...
+  mode: "protect"
+  mutating: false
+  rules:
+    # array as in the CRD spec
+```
+
+For `spec.settings`, depending on the Helm chart it may be a well formed YAML
+object or a multiline string, since the CRD defines it as a [free form
+object](https://docs.kubewarden.io/reference/CRDs#policyspec).
+
```
jordojordo, line 156: We should add a section here outlining the requirement of a `questions.yaml` in the chart's root directory. We can copy the current `questions-ui.yml` that already exists for each policy which requires extra settings (For example: [capabilities-psp-policy questions](https://github.com/kubewarden/capabilities-psp-policy/blob/main/questions-ui.yml))

rfc/0010-rancher-integration-policies.md
viccuad: I would prefer to keep the same names on already existing annotations, and substitute the `spec.X` for `kubewarden/X`, I think it simplifies things and also is forward-compatible if the tools just copy all existing annotations.

viccuad: On second thought, given that we are mirroring the [artifacthub API](https://artifacthub.io/docs/topics/annotations/kubewarden/), I think we should indeed go with your suggestion. It also matches the current artifacthub-pkg.yml. Will amend.

rfc/0010-rancher-integration-policies.md
viccuad: Done.

rfc/0010-rancher-integration-policies.md
viccuad: At the end, I went with the annotations as they are in `metadata.yml`. For the info that isn't present as annotations in `metadata.yml` but as normal keys, we will create Helm chart annotations that mirror those in `artifacthub-pkg.yml`.

I find it better to only need to obtain info from `metadata.yml` instead of `metadata.yml` and `artifacthub-pkg.yml`. This simplifies automation.

rfc/0010-rancher-integration-policies.md
```diff
@@ -47,34 +48,166 @@ Problems to solve:
...
 
-The resulting policy charts will not be served and released by us, but used to
-build Rancher charts from source by using the
-[Package](https://github.com/rancher/charts/blob/dev-v2.6/docs/packages.md)
-format in https://github.com/rancher/charts.
-Policy charts will be submitted via a
-`packages/kubewarden/<policy>.package.yaml`, with `package.yaml::commit`
-pointing to the relevant `rancher-X` branch. Rancher charts are build from those
-`package.yaml` files.
+Even if the charts will not be publitized to users, the repository can be consumed
```
jordojordo:
```suggestion
Even if the charts will not be publicized to users, the repository can be consumed
```

jordojordo (approved):

fabriziosestito (approved):

jvanz (approved):

PR #46, opened 2025-02-11
Use a common Helm template for all policies

viccuad:
## Description

Relates to https://github.com/rancher/kubewarden-ui/issues/1024

Specify how to use the Chart annotations to generate a common Chart policy template/policy.yml.

With this we should have the expected specification complete.

## Test

See POC on https://github.com/kubewarden/helm-charts/compare/main...rancher-policies.

### DIFF
--- a/rfc/0010-rancher-integration-policies.md
+++ b/rfc/0010-rancher-integration-policies.md
@@ -113,6 +113,13 @@ there.
 
 The following Rancher Helm chart annotations must be present:
 
+- `catalog.cattle.io/certified`: rancher
+- `catalog.cattle.io/ui-component` to `kubewarden`: This is added for custom UI deployment of a chart
+- `catalog.cattle.io/os` to `linux`.
+- `catalog.cattle.io/permits-os` to `linux,windows`
+- `catalog.cattle.io/upstream-version` to `"<version of policy chart>"`: The version
+  of the upstream chart or app. It prevents the unexpected "downgrade" when
+  upgrading an installed chart that uses our 100.x.x+upVersion version schema.
 - `catalog.cattle.io/ui-component: kubewarden`. Added for custom UI deployment of a chart.
 - `catalog.cattle.io/hidden` to `"true"`.
 - `catalog.cattle.io/type` to a new type, `kubewarden-policy`.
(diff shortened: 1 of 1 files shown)


flavio (approved):
LGTM

rfc/0010-rancher-integration-policies.md
```diff
@@ -123,8 +146,8 @@ The following Rancher Helm chart annotations must be present:
 
 And the following must be missing:
 
-- `catalog.cattle.io/scope`. Given that we are setting
-  `catalog.cattle.io/hidden` to true, it is not relevant.
+- `catalog.cattle.io/scope`. Given that we are setting `catalog.cattle.io/hidden`
+  to true, it is not relevant.
```
jordojordo, line 151: We need to ensure that these annotations are also applied to every `Chart.yaml`:

```yaml
  kubewarden/displayName: <policy-displayName>
  kubewarden/mutation: 'true|false'
  kubewarden/contextAwareResources: 'true|false'
  kubewarden/resources: <comma-separated-string>
```

Without these the policy table will use default values but will lead to a poor user experience:

Screenshot_20250211_111226

We use the the reference annotations for:
- `kubewarden/displayName`: will show a "pretty printed" name (defaults to `chart.name`
- `kubewarden/mutation` and `kubewarden/contextAwareResources` are used within the "Features" column
- `kubewarden/resources` is used in the "Resource Type"

All of these annotations will also be added into the available options for filtering by "Attribute" as well as input filtering.

They're necessary in the `Chart.yaml` due to the limitations of the Rancher API being unable to fetch the `values` prior to making costly request calls per each chart.

jordojordo (changes requested):

rfc/0010-rancher-integration-policies.md
viccuad: They are there, but with the keys matching the ones in metadata.yml:
https://github.com/kubewarden/rfc/pull/46/files#diff-958ecb40fefeb8fddf465473929b53be370094a241988f9d045ed8f50a8ffca8R108-R125

So, `io.artifacthub.displayName`, `io.artifacthub.resources`, etc. Plus the new `kubewarden/mutation` and `kubewarden/contextAwareResources`.

jordojordo: Okay that works, I can add in some logic for those annotations as well.

jordojordo (approved):
lgtm

PR #51, opened 2025-08-29
feat: Kubewarden stack telemetry.

### LINKED ISSUE #1167: Spike: research how to emit telemetry
We want to collect some data about the usage of Kubewarden.

This issue is about writing an RFC that describes which telemetry data to collect, how to perform the collection and where to send it.

We can ask to the Longhorn developers how they are doing that.

## Acceptance Criteria

Write an RFC that answers these questions:

- [ ] Determine the data that can be collected
- [ ] Take into account the existence of an opt-out flag
- [ ] Determine where to push the data
- [ ] How often the data has to be collected
- [ ] Be respectful of CNCF telemetry guidelines (TODO: insert doc)

jvanz:
## Description

Adds initial draft of the RFC describing how Kubearden can send environment telemetry to central location for further analisys.

Fix https://github.com/kubewarden/kubewarden-controller/issues/1167

### DIFF
--- a/rfc/0023-kubewarden-stack-telemetry.md
+++ b/rfc/0023-kubewarden-stack-telemetry.md
@@ -161,11 +161,11 @@ structure:
 {
   "appVersion": "v1.28.0",
   "extraTagInfo": {
-    "kubernetesVersion": "v1.33.0",
+    "kubernetesVersion": "v1.33.0"
   },
   "extraFieldInfo": {
     "policyServerCount": 2,
-    "policiesCount": 3
+    "policiesCount": 3,
     "namespaceUid": "<namespace-uid-where-kubewarden-is-installed>"
   }
 }
@@ -202,11 +202,14 @@ The same data point will have the following fields:
 - `policiesCount`: The number of policies deployed in the cluster.
 - `namespaceUid`: A UID to uniquely identify the Kubewarden stack installation.
   This should be UID of the namespace where Kubewarden is installed.
-- `value`: A field always set to `1`,
+- `value`: A
+  [field](https://github.com/longhorn/upgrade-responder/blob/a6f6c7736b7e420b07ae7d813765dac778ebc638/upgraderesponder/service.go#L49C36-L49C95)
+  always set to `1`,
   [added](https://github.com/longhorn/upgrade-responder/blob/a6f6c7736b7e420b07ae7d813765dac778ebc638/upgraderesponder/service.go#L560)
-  automatically by `upgrade-responder`. This default value ensures that the
-  data point has a value for metric calculations, even if the request contains no
-  other fields.
+  automatically by `upgrade-responder`. This default value ensures that the data
+  point has a value for metric calculations, even if the request contains no
+  other fields. In other words, it's a dummy InfluxDB field used to count the
+  number of data points
 
   ### Request validation schema
 
--- a/rfc/0023-kubewarden-stack-telemetry.md
+++ b/rfc/0023-kubewarden-stack-telemetry.md
@@ -14,6 +14,9 @@ about user applications, but rather to learn more about the typical
 environments where Kubewarden is deployed. This data will help guide the
 project's future development.
 
+Kubearden stack telemetry will be enabled by default, but users will have an
+easy way to disable it
+
 # Motivation
 
 The Kubewarden team currently has limited visibility into the environments
@@ -113,6 +116,13 @@ This new periodic reconciler will start a time.Ticker that will periodically:
 - Collect the Kubernetes version.
 - Build a JSON payload with the collected data and send it to the remote server
   using secure communication.
+- The controller logs the available updates for the Kubewarden version running returned by
+  the `upgrade-responder` server
+
+The execution of this new reconciler should never interrupt the functionality
+of the remaining reconcilers. This means that if the controller is
+misconfigured with invalid endpoint, the reconciler should log an error and
+continue.
 
 ## Telemetry Server
 
@@ -144,6 +154,8 @@ When the `upgrade-responder` receives a request, it will:
 2. [Infer](https://github.com/longhorn/upgrade-responder/blob/a6f6c7736b7e420b07ae7d813765dac778ebc638/upgraderesponder/service.go#L509)
    the origin's geo-location from the request's source IP address.
 3. Store the enriched data in an InfluxDB database.
+4. Respond the request with the available updates to the given Kubewarden version
+   defined in the `appVersion` field.
 
 The `upgrade-responder` will write the request data into an InfluxDB metric. The
 `extraTagInfo` values will be added as
@@ -209,6 +221,10 @@ stackTelemetry:
 
 When `stackTelemetry.enabled` is `true`, the feature is enabled in the
 controller via the corresponding CLI flags.
+
+The `stackTelemetry.enabled` setting will be `true` by default. Users will be
+notified during the Kubewarden stack upgrade that this telemetry is included,
+and instructions to disable it will be provided
 
 # Drawbacks
 
--- a/rfc/0023-kubewarden-stack-telemetry.md
+++ b/rfc/0023-kubewarden-stack-telemetry.md
@@ -80,7 +80,7 @@ The main changes to the Kubewarden controller would be:
   information.
 - A periodic job within the controller to collect and send the information.
 
-## Controller Configuration
+### Controller Configuration
 
 The controller should have two new CLI flags to enable telemetry collection and
 delivery: `--stack-telemetry-endpoint` and `--stack-telemetry-period`. The first
@@ -94,7 +94,7 @@ manager --leader-elect ... --stack-telemetry-endpoint="https://metrics.kubewarde
 When the `--stack-telemetry-endpoint` flag is not defined, no telemetry will be
 sent. The feature will be disabled by default.
 
-## Controller Information Collection and Delivery
+### Controller Information Collection and Delivery
 
 When the telemetry endpoint is configured, a new periodic job should be created
 in the controller to collect and send the data. This job can be implemented
@@ -103,10 +103,8 @@ into the manager alongside the existing reconcilers.
 
 This new periodic reconciler will start a time.Ticker that will periodically:
 
-- Collect information about how many PolicyServers are deployed, including
-  their versions.
-- Collect information about how many policies are deployed, including the
-  policy modules used.
+- Collect information about how many PolicyServers are deployed
+- Collect information about how many policies are deployed
 - Collect the Kubewarden version in use (this can be a constant updated on
   every release).
 - Collect the Kubernetes version.
@@ -115,59 +113,89 @@ This new periodic reconciler will start a time.Ticker that will periodically:
 
 ## Telemetry Server
 
-The telemetry server is a custom application that receives information from
-controllers and store it into a database for further analysis. This server
-will expose an endpoint to receive a JSON payload with the following structure:
+The telemetry server will be an instance of the [Longhorn
+upgrade-responder](https://github.com/longhorn/upgrade-responder).
+This server will receive information from controllers and store it in an
+InfluxDB database for further analysis.
 
-```
+The server exposes an endpoint to receive a JSON payload with the following
+structure:
+
+```json
 {
-  "instanceUid": "unique-id-for-the-cluster",
-  "kubewardenVersion": "v1.28.0",
-  "kubernetesVersion": "v1.33.0",
-  "policyServers": [
-    {
-      "uid": "policy-server-uid",
-      "imageDigest": "sha256:6c7d6433524e1dd7def937e123ee8f8f8da993fdc2297efaa6640b44f3285eee",
-      "policies": [
-        { "moduleDigest": "sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a" },
-        { "moduleDigest": "sha256:eef38c001a43600b2fb7e7d0609c8f26fbd93d8142476406988d0374d36d0bfc" }
-      ]
-    }
-  ]
+  "appVersion": "v1.28.0",
+  "extraTagInfo": {
+    "kubernetesVersion": "v1.33.0",
+    "instanceUid": "<unique-id-for-controller>"
+  },
+  "extraFieldInfo": {
+    "policyServerCount": 2,
+    "policiesCount": 3
+  }
 }
-
 ```
 
-In this payload, container images and policy modules are identified by their
-manifest digest to avoid exposing metadata about internal environments (like
-private registries) and to prevent confusion from different tags pointing to
-the same OCI artifact. However, this implies that the telemetry reconciler must
-fetch this information from the OCI registry at runtime.
+When the `upgrade-responder` receives a request, it will:
 
-When the telemetry server receives the request, it will:
+1. Validate the payload using a predefined JSON schema.
+2. [Infer](https://github.com/longhorn/upgrade-responder/blob/a6f6c7736b7e420b07ae7d813765dac778ebc638/upgraderesponder/service.go#L509)
+   the origin's geo-location from the request's source IP address.
+3. Store the enriched data in an InfluxDB database.
 
-1. Validate the payload.
-2. Infer the origin location from the source IP address.
-3. Look up policy and PolicyServer versions from the OCI artifact digest.
-4. Store the enriched data in a time-series database.
+The `upgrade-responder` will write the request data into an InfluxDB metric. The
+`extraTagInfo` values will be added as
+[tags](https://github.com/longhorn/upgrade-responder/blob/a6f6c7736b7e420b07ae7d813765dac778ebc638/upgraderesponder/service.go#L538),
+while the `extraFieldInfo` values will be added as
+[fields](https://github.com/longhorn/upgrade-responder/blob/a6f6c7736b7e420b07ae7d813765dac778ebc638/upgraderesponder/service.go#L558).
 
-The telemetry server should store data to the Prometheus time series database
-to be used later as a source for a Grafana dashboard. The metric should something like
-the following:
+For example, the JSON payload above would result in a data point with the
+following tags:
 
-```
-deployed-policies{	"policy_server_id"="uid",
-	"policy_server_image"="sha256:6c7d6433524e1dd7def937e123ee8f8f8da993fdc2297efaa6640b44f3285eee",
-	"module"="sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a",
-	"kubewarden_version"="v1.28.0",
-	"instance-uid"="uid"
-	"country"="<some country>",
-	"city"="<some city>"
-```
+- `appVersion`: The Kubewarden version (e.g., v1.28.0).
+- `kubernetesVersion`: The Kubernetes version where the controller is running
+  (e.g., v1.33.0).
+- `instanceUid`: A UID to uniquely identify the controller instance.
+- `city`: City name extracted from the request's source IP.
+- `country`: Country name extracted from the request's source IP.
+- `country_isocode`: Country ISO code extracted from the request's source IP.
+
+The same data point will have the following fields:
+
+- `policyServerCount`: The number of PolicyServers deployed in the cluster.
+- `policiesCount`: The number of policies deployed in the cluster.
+- `value`: A field always set to `1`,
+  [added](https://github.com/longhorn/upgrade-responder/blob/a6f6c7736b7e420b07ae7d813765dac778ebc638/upgraderesponder/service.go#L560)
+  automatically by `upgrade-responder`.
+
+### Understanding Tags and Fields
+
+There are differences from InfluxDB `tags` and `fields` which may be relevant
+to document here as well. Tags are indexed metadata which can be used to group
+and filter data points. While `fields` are the actual metric data.
+
+| Aspect          | Tags                                                                                | Fields                                                                |
+| :-------------- | :---------------------------------------------------------------------------------- | :-------------------------------------------------------------------- |
+| **Purpose**     | Serve as indexed metadata used for filtering and grouping (e.g., `host`, `region`). | Hold the actual measured values (e.g., `temperature`, `cpu_usage`).   |
+| **Indexing**    | Stored in an inverted index; queries on tags are very fast.                         | Not indexed; queries must scan data, which can be slower.             |
+| **Cardinality** | High cardinality (many unique values) can degrade database performance.             | Low impact on performance; many distinct field values are acceptable. |
+| **Data Types**  | Limited to strings.                                                                 | Support numeric types, booleans, and strings.                         |
+
+### Grafana Dashboard
+
+Alongside the `upgrade-responder` and InfluxDB, a Grafana instance with a custom
+dashboard will be configured. The dashboard will use InfluxDB as its data
+source and provide queries to visualize the desired metrics.
+
+### Telemetry Server Infrastructure
+
+The `upgrade-responder`, InfluxDB, and Grafana instances used for Kubewarden
+telemetry will be maintained by the SUSE team. All members of the Kubewarden
+organization on GitHub will have access to the collected data.
 
 ## Helm Chart Changes
 
-The Kubewarden Helm charts would require new values to configure the controller's CLI flags.
+The Kubewarden Helm charts would require new values to configure the
+controller's CLI flags.
 
 ```yaml
 stackTelemetry:
@@ -181,20 +209,18 @@ controller via the corresponding CLI flags.
 
 # Drawbacks
 
-- Requires building and maintaining a custom client, a custom server, and a
-  custom data pipeline.
-- When new metadata is required, it necessitates code changes and a new release
-  for both the controller and the telemetry server. The JSON schema versions must
-  be kept in sync.
+- When new metadata or more metric data is required, it necessitates code
+  changes and a new release for both the controller and, maybe, the telemetry
+  server. The JSON schema versions must be kept in sync.
 - We are solely responsible for ensuring secure communication (TLS,
   authentication, etc.) with the remote telemetry server.
 
 # Alternatives
 
-If we proceed with the OpenTelemetry integration, we should investigate if the
-existing metrics can be updated to achieve the same goal. It might be possible
-to reuse current metrics and simply add a new Otel collector pipeline to send
-them to the remote collector, minimizing controller changes.
+The following subsection is a option given during the write of this RFC. But it
+was not select by the majority of the team. The team decided to move with the
+`upgrade-responder` option because the people that which will maintain the
+infrastructure more familiarized with the tool.
 
 ## Not selected option: OpenTelemetry Integration
 
@@ -324,3 +350,5 @@ to add location information to the received metrics.
 [opentelemetry-collector-contrib/processor at main · open-telemetry/opentelemetry-collector-contrib · GitHub](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor)
 
 [opentelemetry-collector/processor at main · open-telemetry/opentelemetry-collector · GitHub](https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor)
+
+https://github.com/longhorn/upgrade-responder
--- a/rfc/0023-kubewarden-stack-telemetry.md
+++ b/rfc/0023-kubewarden-stack-telemetry.md
@@ -4,7 +4,7 @@
 | Start Date   | 2025-08-28                                |
 | Category     | telemetry                                 |
 | RFC PR       | https://github.com/kubewarden/rfc/pull/51 |
-| State        | **TO-BE-DEFINED**                         |
+| State        | **IN-REVIEW**                             |
 
 # Summary
 
@@ -49,8 +49,11 @@ introduce a significant security vulnerability into my environment.
 As a Kubewarden user, I want to ensure that the process of sending telemetry
 data does not consume excessive resources in my environment.
 
-As a Kubewarden user, I want an easy way to disable the reconciler which send
-environment telemetry data to a remote central location.
+As a Kubewarden user, I want an easy way to disable the Kubewarden environment
+telemetry data sending
+
+As a Kubewarden user, I want to receive warning in the logs about available
+upgrades
 
 # Detailed design
 
@@ -109,20 +112,40 @@ into the manager alongside the existing reconcilers.
 
 This new periodic reconciler will start a time.Ticker that will periodically:
 
-- Collect information about how many PolicyServers are deployed
-- Collect information about how many policies are deployed
+- Collect information about how many PolicyServers are deployed.
+- Collect information about how many policies are deployed.
 - Collect the Kubewarden version in use (this can be a constant updated on
   every release).
 - Collect the Kubernetes version.
 - Build a JSON payload with the collected data and send it to the remote server
   using secure communication.
 - The controller logs the available updates for the Kubewarden version running returned by
-  the `upgrade-responder` server
+  the [upgrade-responder](https://github.com/longhorn/upgrade-responder) server.
 
 The execution of this new reconciler should never interrupt the functionality
 of the remaining reconcilers. This means that if the controller is
 misconfigured with invalid endpoint, the reconciler should log an error and
 continue.
+
+## Helm Chart Changes
+
+The Kubewarden Helm charts would require new values to configure the
+controller's CLI flags.
+
+```yaml
+stackTelemetry:
+  enabled: true
+  endpoint: "https://metrics.kubewarden.io"
+  period: "1h"
+```
+
+When `stackTelemetry.enabled` is `true`, the feature is enabled in the
+controller via the corresponding CLI flags.
+
+The `stackTelemetry.enabled` setting will be `true` by default. Users will be
+notified during the Kubewarden stack install and upgrade via Helm NOTES.txt
+that this telemetry is included, and instructions to disable it will be
+provided
 
 ## Telemetry Server
 
@@ -180,7 +203,9 @@ The same data point will have the following fields:
 - `policiesCount`: The number of policies deployed in the cluster.
 - `value`: A field always set to `1`,
   [added](https://github.com/longhorn/upgrade-responder/blob/a6f6c7736b7e420b07ae7d813765dac778ebc638/upgraderesponder/service.go#L560)
-  automatically by `upgrade-responder`.
+  automatically by `upgrade-responder`. This default value ensures that the
+  data point has a value for metric calculations, even if the request contains no
+  other fields.
 
 ### Understanding Tags and Fields
 
@@ -207,39 +232,20 @@ The `upgrade-responder`, InfluxDB, and Grafana instances used for Kubewarden
 telemetry will be maintained by the SUSE team. All members of the Kubewarden
 organization on GitHub will have access to the collected data.
 
-## Helm Chart Changes
-
-The Kubewarden Helm charts would require new values to configure the
-controller's CLI flags.
-
-```yaml
-stackTelemetry:
-  enabled: true
-  endpoint: "https://metrics.kubewarden.io"
-  period: "1h"
-```
-
-When `stackTelemetry.enabled` is `true`, the feature is enabled in the
-controller via the corresponding CLI flags.
-
-The `stackTelemetry.enabled` setting will be `true` by default. Users will be
-notified during the Kubewarden stack upgrade that this telemetry is included,
-and instructions to disable it will be provided
-
 # Drawbacks
 
 - When new metadata or more metric data is required, it necessitates code
   changes and a new release for both the controller and, maybe, the telemetry
   server. The JSON schema versions must be kept in sync.
-- We are solely responsible for ensuring secure communication (TLS,
-  authentication, etc.) with the remote telemetry server.
 
 # Alternatives
 
 The following subsection is a option given during the write of this RFC. But it
 was not select by the majority of the team. The team decided to move with the
-`upgrade-responder` option because the people that which will maintain the
-infrastructure more familiarized with the tool.
+`upgrade-responder` option because this is a solution used by other CNCF
+projects (Longhorn) and we (SUSE Rancher - who is willing to pay for the
+infrastructure) also have experience managing and maintaining the server-side
+infrastructure
 
 ## Not selected option: OpenTelemetry Integration
 
--- a/rfc/0023-kubewarden-stack-telemetry.md
+++ b/rfc/0023-kubewarden-stack-telemetry.md
@@ -56,6 +56,17 @@ This feature involves creating a dedicated controller reconciler
 that sends telemetry data to a remote server. It requires changes in some
 Kubewarden components, as described below.
 
+## Desired metrics
+
+The changes described in this document aim to allow the Kubewarden team to
+gather the following metrics:
+
+- Number of active PolicyServers
+- Number of active policies
+- Average number of policies per PolicyServer
+- Number of Kubewarden installations by version
+- Number of Kubewarden installations by Kubernetes version
+
 ## Controller Changes
 
 To send information about the environment, the controller would be responsible
--- a/rfc/0023-kubewarden-stack-telemetry.md
+++ b/rfc/0023-kubewarden-stack-telemetry.md
@@ -9,9 +9,9 @@
 # Summary
 
 This RFC describes how the Kubewarden stack will send telemetry data about the
-environment in which it is running. The goal is not to collect information
-about user applications, but rather to learn more about the typical
-environments where Kubewarden is deployed. This data will help guide the
+environment in which it is running. The goal isn't to collect information
+about user applications, but rather to learn more about typical Kubewarden deployment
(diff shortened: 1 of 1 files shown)


rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,338 @@
+|              |                                  |
+| :----------- | :------------------------------- |
+| Feature Name | Kubewarden environment telemetry |
+| Start Date   | 2025-08-28                       |
+| Category     | telemetry                        |
+| RFC PR       | [fill this in after opening PR]  |
```
flavio: This can be filled now

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,338 @@
...
+As a Kubewarden user, I want to help improve the project by providing anonymous
+usage data without compromising my privacy or violating data protection
+regulations.
+
+As a Kubewarden user, I want to ensure that sending telemetry data does not
+introduce a significant security vulnerability into my environment.
+
+As a Kubewarden user, I want to ensure that the process of sending telemetry
+data does not consume excessive resources in my environment.
+
```
flavio, line 51: Maybe we can add a user story about a user who doesn't want telemetry data to be sent at all.

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,338 @@
...
+| State        | **TO-BE-DEFINED**                |
+
+# Summary
+
+This RFC describes how the Kubewarden stack will send telemetry data about the
+environment in which it is running. The goal is not to collect information
+about user applications, but rather to learn more about the typical
+environments where Kubewarden is deployed. This data will help guide the
+project's future development.
+
```
flavio, line 16: I would add a paragraph stressing out this can be turned off

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,338 @@
...
+This section is also split into two subsections, one for each implementation
+option.
+
+## OpenTelemetry Integration
+
+- Potential for breaking changes in the OpenTelemetry specification or
+  collector components.
+- If a user already has a custom OpenTelemetry collector, integrating our
+  telemetry pipeline might be complex or they might filter our data out.
```
flavio: I think we can run as many collectors as we want. Meaning, that we could always deploy our own collector that sends the telemetry data to the remote location, regardless of the user having its own collector.

flavio (commented):
Thanks for having written this RFC.

~~I think the best option would be to go with the OpenTelemetry approach.~~

I've changed my mind, I think the best option is to go with the second approach since it would allow us to reuse the same technology Longhorn uses.

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,338 @@
...
+
+This section describes two options to achieve this goal. Therefore, this
+section is split into two subsections: the first option using OpenTelemetry,
+and the second creating a dedicated controller reconciler.
+
+> [!WARNING]
+> As a CNCF project, it is necessary to get approval from the CNCF to implement either of the following options.
+> [Telemetry Data Collection and Usage Policy | Linux Foundation](https://www.linuxfoundation.org/legal/telemetry-data-policy)
+
+## Option 1: OpenTelemetry Integration
```
olblak: My main concern with this approach, is that we expect end users to provide extra effort to provide telemetry, which won't be done by most of end-users. If we don't opt in by default telemetry then we need to be sure that it can be enabled in the most straightforward way.

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,338 @@
...
+```go
+policiesGauge, err := meter.Int64Gauge("deployed-policies", metric.WithDescription("The number of deployed policies"))
+if err != nil {
+    log.Fatal(err)
+}
+// Attributes represent additional key-value descriptors that can be bound
+// to a metric observer or recorder.
+commonAttrs := []attribute.KeyValue{
+    attribute.String("policy_server_id", policy.GetPolicyServer().GetUID()),
+    attribute.String("policy_server_image", policy.GetPolicyServer().GetImageDigest()),
```
olblak, line 323: Are we planning to store policy names? This could lead to end user data leak depending on the name 
for example:

`oci://registry.corporate.com/mysupersecretpolicy`

olblak: Re-reading my comment, I realize that `policy_server_image` is for the server image but still what if the end user rebuilds the kubewarden images, is that an information we should track here

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,338 @@
...
+
+```yaml
+receivers:
+  otlp:
+    protocols:
+      grpc: {}
+processors:
+  # Control how much time we can wait before sending the metric to the remote Otel collector
+  batch:
+    timeout: 1h
```
olblak, line 351: If we send the metrics every hour then it sounds more like real time monitoring

rfc/0023-kubewarden-stack-telemetry.md
olblak: It depends if we want to opt-in or opt-out by default.
As a project maintainer I would prefer to opt-in by default, it provide more data
As a project user, I would prefer to opt out by default :)

Ultimately I am fine with both but we need to take a decision on this

rfc/0023-kubewarden-stack-telemetry.md
jvanz: Answering the question, the metric should not contains the policy server name or the policy name. To do not expose the registry url and other local information by accident. This proposal, which was not select by the team, would send only the oci artifact digest. No names. But this is not selected integration option.

rfc/0023-kubewarden-stack-telemetry.md
jvanz: This is a example configuration. It can be bigger. I just use the Longhorn [example](https://github.com/longhorn/upgrade-responder?tab=readme-ov-file#how-it-works). 

> Server will expect each instance of the application makes a request once per hour.

rfc/0023-kubewarden-stack-telemetry.md
jvanz: Yes. Considering that we allow users to use their own [custom Otel collector](https://docs.kubewarden.io/howtos/telemetry/custom-otel-collector#install-kubewarden-stack),  this would require the controller to send metrics to two different Otel collector. 

But as Otel integration will not be done, this is not relevant anymore.

davideiori1:
Hi.  I suggest adding a new metric to list what are the policies in use. This is to enable us to rank the most used policies vs the least used ones. This should be done only for the default policies provided by Kubewarden so that we do not track the names of the custom policies created by users.

rfc/0023-kubewarden-stack-telemetry.md
viccuad: Reopening, we can set the state to IN-REVIEW or so, while we wait for CNCF approval.

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,373 @@
...
+
+# Summary
+
+This RFC describes how the Kubewarden stack will send telemetry data about the
+environment in which it is running. The goal is not to collect information
+about user applications, but rather to learn more about the typical
+environments where Kubewarden is deployed. This data will help guide the
+project's future development.
+
+Kubearden stack telemetry will be enabled by default, but users will have an
```
viccuad:
```suggestion
Kubewarden stack telemetry will be enabled by default, but users will have an
```

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,373 @@
...
+easy way to disable it
```
viccuad:
```suggestion
easy way to disable it.
```

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,373 @@
...
+
+# Motivation
+
+The Kubewarden team currently has limited visibility into the environments
+where users deploy the project. This information is invaluable for making
+data-driven decisions about the project's future. For this reason, we would
+like to add a feature to the Kubewarden stack that allows sending
+non-sensitive, anonymous environment information to a central service for
+aggregation and analysis
```
viccuad:
```suggestion
aggregation and analysis.
```

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,373 @@
...
+usage data without compromising my privacy or violating data protection
+regulations.
+
+As a Kubewarden user, I want to ensure that sending telemetry data does not
+introduce a significant security vulnerability into my environment.
+
+As a Kubewarden user, I want to ensure that the process of sending telemetry
+data does not consume excessive resources in my environment.
+
+As a Kubewarden user, I want an easy way to disable the reconciler which send
```
viccuad: I would reword and don't mention the reconciler nor a central location, just "an easy way to disable sending telemetry data".
The rest of user stories don't mention the reconciler, we haven't presented it yet, nor it is needed for understanding the story.

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,373 @@
...
+### Controller Information Collection and Delivery
+
+When the telemetry endpoint is configured, a new periodic job should be created
+in the controller to collect and send the data. This job can be implemented
+using the Runnable interface from the controller-runtime package and integrated
+into the manager alongside the existing reconcilers.
+
+This new periodic reconciler will start a time.Ticker that will periodically:
+
+- Collect information about how many PolicyServers are deployed
```
viccuad: There's a mix of phrases ending in dots and missing them throughout the document, I would like to have consistency.

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,373 @@
...
+- Collect information about how many policies are deployed
+- Collect the Kubewarden version in use (this can be a constant updated on
+  every release).
+- Collect the Kubernetes version.
+- Build a JSON payload with the collected data and send it to the remote server
+  using secure communication.
+- The controller logs the available updates for the Kubewarden version running returned by
+  the `upgrade-responder` server
```
viccuad: Here is the first mention of the upgrade-responder feature. I would like to mention in the user stories that the user would get a (configurable) notification of new Kubewarden releases.

Edit:
this is the first and only mention, as it isn't listed on the controller nor helm chart changes, nor explained how the user would receive notifications. Maybe is a left over?

Since we expect users to manage the Kubewarden stack via Gitops, maybe this feature is not of much value, and can be dropped off.

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,373 @@
...
+while the `extraFieldInfo` values will be added as
+[fields](https://github.com/longhorn/upgrade-responder/blob/a6f6c7736b7e420b07ae7d813765dac778ebc638/upgraderesponder/service.go#L558).
+
+For example, the JSON payload above would result in a data point with the
+following tags:
+
+- `appVersion`: The Kubewarden version (e.g., v1.28.0).
+- `kubernetesVersion`: The Kubernetes version where the controller is running
+  (e.g., v1.33.0).
+- `instanceUid`: A UID to uniquely identify the controller instance.
```
viccuad: I believe we need to explain how this `instanceUid` is unique. Is it unique per kubewarden-controller Deployment, or Kubewarden stack release (helm release)? 

I would expect it to be the second one. This allows us to not get duplicated data from several kubewarden-controller Deployments or restarts, and keeps consistent throughout upgrades.
Hence I believe it should be a random UID value created on `helm install`, not overwritten with `helm upgrade`, and saved as an opaque value in the Helm release ConfigMaps or Secrets.

We should also mention it in its relevant section (for example "Helm chart changes").

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,373 @@
...
+- `city`: City name extracted from the request's source IP.
+- `country`: Country name extracted from the request's source IP.
+- `country_isocode`: Country ISO code extracted from the request's source IP.
+
+The same data point will have the following fields:
+
+- `policyServerCount`: The number of PolicyServers deployed in the cluster.
+- `policiesCount`: The number of policies deployed in the cluster.
+- `value`: A field always set to `1`,
```
viccuad: Could we explain why the field is always set to `1`? (looked at the upgrade-responder but it was not inmediately obvious, and either the reader should know why, or not care why).

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,373 @@
...
+dashboard will be configured. The dashboard will use InfluxDB as its data
+source and provide queries to visualize the desired metrics.
+
+### Telemetry Server Infrastructure
+
+The `upgrade-responder`, InfluxDB, and Grafana instances used for Kubewarden
+telemetry will be maintained by the SUSE team. All members of the Kubewarden
+organization on GitHub will have access to the collected data.
+
+## Helm Chart Changes
```
viccuad: I would move this section upwards, after the "Controller changes" section. Hence one has the full changes in mind, to read how the infrastructure will interact with them.

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,373 @@
...
+| **Purpose**     | Serve as indexed metadata used for filtering and grouping (e.g., `host`, `region`). | Hold the actual measured values (e.g., `temperature`, `cpu_usage`).   |
+| **Indexing**    | Stored in an inverted index; queries on tags are very fast.                         | Not indexed; queries must scan data, which can be slower.             |
+| **Cardinality** | High cardinality (many unique values) can degrade database performance.             | Low impact on performance; many distinct field values are acceptable. |
+| **Data Types**  | Limited to strings.                                                                 | Support numeric types, booleans, and strings.                         |
+
+### Grafana Dashboard
+
+Alongside the `upgrade-responder` and InfluxDB, a Grafana instance with a custom
+dashboard will be configured. The dashboard will use InfluxDB as its data
+source and provide queries to visualize the desired metrics.
```
viccuad, line 202: Do we intend to have this dashboard public, or only accessible to Kubewarden maintainers?
If the second one (the usual one), I believe it would be good to document this here, and how the login is managed, as this seems to be part of the CNCF legal review.

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,373 @@
...
+
+### Telemetry Server Infrastructure
+
+The `upgrade-responder`, InfluxDB, and Grafana instances used for Kubewarden
+telemetry will be maintained by the SUSE team. All members of the Kubewarden
+organization on GitHub will have access to the collected data.
```
viccuad: I would like to be specific here, maybe even linking to the [CONTRIBUTOR_LADDER.md](https://github.com/kubewarden/community/blob/main/CONTRIBUTOR_LADDER.md). Should it be only Kubewarden maintainers?

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,373 @@
...
+  enabled: true
+  endpoint: "https://metrics.kubewarden.io"
+  period: "1h"
+```
+
+When `stackTelemetry.enabled` is `true`, the feature is enabled in the
+controller via the corresponding CLI flags.
+
+The `stackTelemetry.enabled` setting will be `true` by default. Users will be
+notified during the Kubewarden stack upgrade that this telemetry is included,
```
viccuad:
```suggestion
notified during the Kubewarden stack install and upgrade via Helm NOTES.txt that this telemetry is included,
```

viccuad (commented):
Thanks for the RFC!
I generally like it; I would like to clarify some points (the `instanceUID`) and access to processed data, plus I have some style and ordering quibbles though.

rfc/0023-kubewarden-stack-telemetry.md
flavio: I agree with Victor, the reconciler is a technical implementation, what the user cares is the end result: have an easy way to opt-out form telemetry

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,373 @@
...
+
+The changes described in this document aim to allow the Kubewarden team to
+gather the following metrics:
+
+- Number of active PolicyServers
+- Number of active policies
+- Average number of policies per PolicyServer
+- Number of Kubewarden installations by version
+- Number of Kubewarden installations by Kubernetes version
+
```
flavio, line 79: I think it would be useful to have statistics about the policies being used. We should emit statistics only for the policies that have been authored by the Kubewarden project. We should not track policies created by the users, because that might expose some sensitive details.

We could do something simple, like sending the data only if the policy has been downloaded from the OCI registry of the Kubewarden project.

rfc/0023-kubewarden-stack-telemetry.md
flavio: Also, add a link to the `upgrade-responder` open source project, I know this is mentioned on the next section but it won't harm to have it here too (given it's the 1st time the reader jumps into that)

rfc/0023-kubewarden-stack-telemetry.md
flavio: Good suggestion

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,373 @@
...
+notified during the Kubewarden stack upgrade that this telemetry is included,
+and instructions to disable it will be provided
+
+# Drawbacks
+
+- When new metadata or more metric data is required, it necessitates code
+  changes and a new release for both the controller and, maybe, the telemetry
+  server. The JSON schema versions must be kept in sync.
+- We are solely responsible for ensuring secure communication (TLS,
+  authentication, etc.) with the remote telemetry server.
```
flavio: Is that a drawback? I think this implicit with all telemetry solutions.

The TLS certificate used by the server is going to be issued by a trusted CA that is part of the standard `ca-certificates` bundles.

As for the authorization, this must be in place only for the grafana dashboard, but this will delegate authentication to GitHub

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,373 @@
...
+
+# Alternatives
+
+The following subsection is a option given during the write of this RFC. But it
+was not select by the majority of the team. The team decided to move with the
+`upgrade-responder` option because the people that which will maintain the
+infrastructure more familiarized with the tool.
```
flavio:
```suggestion
`upgrade-responder` option because this is a solution used by other
CNCF projects (Longhorn) and we (SUSE Rancher - who is willing to
pay for the infrastructure) also have experience managing and
maintaining the server-side infrastructure.
```

flavio (approved):
I like the proposal. I've left some minor comments

flavio:
@l-mb: can you please take a look at our plan and educate us about what we would need to do to integrate with [scarf](https://about.scarf.sh/)?

rfc/0023-kubewarden-stack-telemetry.md
jvanz: If we remove the warning from the logs, the reconciler becomes unnecessary. Its goal is to provide users with useful information when they share telemetry data. If the reconciler does nothing with the response about available upgrades, there is no reason to include it. It would just be sending data with no clear benefit to the user. In that case, users could simply query the upgrade-responder from another application without sending metrics in the process

rfc/0023-kubewarden-stack-telemetry.md
jvanz: I'm not sure if this should be only for maintainers. The CNCF folks should have access to this information as well. In the end of the day, they are the "owners" of the project. Therefore, they should have access to this information.

rfc/0023-kubewarden-stack-telemetry.md
jvanz: That's left over from the comparison with the Otel approach. Where we just need to do a configuration change instead of coding that in the controller.

rfc/0023-kubewarden-stack-telemetry.md
jvanz: Yes, the idea was to use to identify kubewarden stack installation. 

I don't like the idea to have an opaque value created by helm. It's solve the issu. But, to me, it looks a too complicated solution for a simple uid generation.  I was reviewing the longhorn code and I have an idea. In the longhord request sending the telemetry data, it sends the longhorn [namespace uid](https://github.com/longhorn/longhorn/blob/9af0bd566dc222cf54055193d119305c0f79496e/dev/upgrade-responder/install.sh#L266). That's solve our issue here as well. It's a unique UID for each Kubewarden installation. And, we do not need to add another resource in our Helm chart templates.

rfc/0023-kubewarden-stack-telemetry.md
jvanz: Good point. We can definitely implement this, but we need to be careful with the data model to avoid performance issues in InfluxDB over time.

The core challenge is tracking a dynamic list of official policies. Storing the names as a single concatenated string in a field ("policy1,policy2") or as individual fields with boolean values ("policy1": true) could make querying difficult and inefficient. Since fields aren't indexed, we'd need to do slow, full data scans to get any meaningful insights. Considering these approaches, these could be examples of the requests payload:


```json
{
  "appVersion": "v1.28.0",
  "extraTagInfo": {
    "kubernetesVersion": "v1.33.0",
  },
  "extraFieldInfo": {
    "policyServerCount": 2,
    "policiesCount": 3,
    "policiesInUse": "ghcr.io/kubewarden/policies/pod-privileged:latest,ghcr.io/kubewarden/policies/user-group-psp:v1.5.0",
    "namespaceUid": "<namespace-uid-where-kubewarden-is-installed>"
  }
```

Or

```json
{
  "appVersion": "v1.28.0",
  "extraTagInfo": {
    "kubernetesVersion": "v1.33.0",
  },
  "extraFieldInfo": {
    "policyServerCount": 2,
    "policiesCount": 3,
    "ghcr.io/kubewarden/policies/pod-privileged:latest": true,
    "ghcr.io/kubewarden/policies/user-group-psp:v1.5.0": true,
    "namespaceUid": "<namespace-uid-where-kubewarden-is-installed>"
  }
```

A better approach would be to add each policy name as a tag. This would allow for fast filtering and grouping. However, this could lead to a high series cardinality problem. Every unique combination of tags creates a new series. As users add or remove official policies, new series are created, leaving old ones dormant. Over time, this could lead to high memory usage and degraded performance. Again, some example:

```json
{
  "appVersion": "v1.28.0",
  "extraTagInfo": {
    "kubernetesVersion": "v1.33.0",
    "ghcr.io/kubewarden/policies/pod-privileged:latest": "true",
    "ghcr.io/kubewarden/policies/user-group-psp:v1.5.0": "true",
  },
  "extraFieldInfo": {
    "policyServerCount": 2,
    "policiesCount": 3,
    "namespaceUid": "<namespace-uid-where-kubewarden-is-installed>"
  }
```

One alternative approach would be to send in the metrics the number of official policies in use. Like this:

```json
{
  "appVersion": "v1.28.0",
  "extraTagInfo": {
    "kubernetesVersion": "v1.33.0",
  },
  "extraFieldInfo": {
    "policyServerCount": 2,
    "policiesCount": 3,
    "officialPoliciesCount": 2,
    "namespaceUid": "<namespace-uid-where-kubewarden-is-installed>"
  }

```

With this approach, we loose the ability to know the individual policies in use. But we can know, at least, the amount of them related to the number of total policies deployed.

jvanz: Another ideia is to have a dedicated metric to keep track of  policy in use. 

```json
{
  "appVersion": "v1.28.0",
  "extraTagInfo": {
    "policyName": "ghcr.io/kubewarden/policies/pod-privileged:latest"
  },
  "extraFieldInfo": {  }
```

That way, we reduce the number of tag in the metrics, witch mitigate the  cardinality problem. And we do not loose the ability to group and filter the data. However, this required another instance of `upgrade-responder` and more request to be sent by the controller (1 for generic metrics + 1 for each policy)

jvanz: If we want to rank the policies as sad by @davideiori1 , I believe we should go with the last option (dedicated metric) or the second option (the one of using policy names in tags but with the possible high cardinality problem)

davideiori1: @jvanz this is indeed a request from my side but I do not want to introduce any major refactoring to accomplish that. If that is the case, I suggest **we keep it simple** for now and record my request as backlog item for future enhancements.

rfc/0023-kubewarden-stack-telemetry.md
viccuad: That sounds good enough to me. I suspect Helm creates a UID per release too, as it needs to know when to upgrade them. But then we would need to configure our charts with RBAC to read the Helm Secrets or ConfigMaps. Your suggestion sounds better indeed.

rfc/0023-kubewarden-stack-telemetry.md
viccuad: Then, we should mention this here. Maybe "people with access to the cncf-kubewarden-maintainers at lists.cncf.io" is the set we want, as that includes both.

rfc/0023-kubewarden-stack-telemetry.md
viccuad: I'm ok keeping the "update notification" feature. But then, I'm not sure if the controller logs is a nice medium to serve that notification.

l-mb:
@flavio I have a pending PR for `longhorn/upgrade-responder` that'd allow us to then mirror incoming upgrade-responder events to said analytics platform as part of the ingestion pipeline, with no changes to how the upgrade-responder clients interact.

So having kubewarden adopt that deployment pattern makes a lot of sense and would fit neatly with the framework. (Scarf.sh is a CNCF-approved solution.)

rfc/0023-kubewarden-stack-telemetry.md
flavio: I think we don't need to argue about the low level format we will use to expose this information, this can be discussed at a later stage when the actual development will start.

What is important here is to agree on the data we plan to send and get an approval for that.

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,413 @@
...
+| RFC PR       | https://github.com/kubewarden/rfc/pull/51 |
+| State        | **IN-REVIEW**                             |
+
+# Summary
+
+This RFC describes how the Kubewarden stack will send telemetry data about the
+environment in which it is running. The goal is not to collect information
+about user applications, but rather to learn more about the typical
+environments where Kubewarden is deployed. This data will help guide the
+project's future development.
```
jhkrug:
```suggestion
This RFC describes how the Kubewarden stack will send telemetry data about the
environment in which it is running. The goal isn't to collect information
about user applications, but rather to learn more about typical Kubewarden deployment
environments. This data will help guide the
project's future development.
```

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,413 @@
...
+As a Kubewarden user, I want an easy way to disable the Kubewarden environment
+telemetry data sending
+
+As a Kubewarden user, I want to receive warning in the logs about available
+upgrades
+
+# Detailed design
+
+> [!WARNING]
+> As a CNCF project, it is necessary to get approval from the CNCF to implement either of the following options.
```
jhkrug:
```suggestion
> As a CNCF project, it's necessary to get approval from the CNCF to implement either of the following options.
```

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,413 @@
...
+## Controller Changes
+
+To send information about the environment, the controller would be responsible
+for collecting and transmitting the data to a remote, central place. The
+controller already has access to all Kubewarden resources and can be granted
+permissions to access other relevant information.
+
+The main changes to the Kubewarden controller would be:
+
+- A configuration option to specify the endpoint that will receive the
```
jhkrug:
```suggestion
- A configuration option to specify the endpoint that receives the
```

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,413 @@
...
+  information.
+- A periodic job within the controller to collect and send the information.
+
+### Controller Configuration
+
+The controller should have two new CLI flags to enable telemetry collection and
+delivery: `--stack-telemetry-endpoint` and `--stack-telemetry-period`. The first
```
jhkrug:
```suggestion
delivery: `--stack-telemetry-endpoint` and `--stack-telemetry-period`. The former
```

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,413 @@
...
+contains the full URL of the remote server. The latter defines the interval at
+which the controller sends data, defaulting to `1h`. For example:
+
+```
+manager --leader-elect ... --stack-telemetry-endpoint="https://metrics.kubewarden.io/" --stack-telemetry-period=6h ...
+```
+
+When the `--stack-telemetry-endpoint` flag is not defined, no telemetry will be
```
jhkrug:
```suggestion
When the `--stack-telemetry-endpoint` flag isn't defined, no telemetry will be
```

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,413 @@
...
+sent. The feature will be disabled by default.
+
+### Controller Information Collection and Delivery
+
+When the telemetry endpoint is configured, a new periodic job should be created
```
jhkrug:
```suggestion
When configuring the telemetry endpoint, a new periodic job should be created
```

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,413 @@
...
+in the controller to collect and send the data. This job can be implemented
+using the Runnable interface from the controller-runtime package and integrated
+into the manager alongside the existing reconcilers.
+
+This new periodic reconciler will start a time.Ticker that will periodically:
```
jhkrug:
```suggestion
This new periodic reconciler will start a `time.Ticker` that will periodically:
```

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,413 @@
...
+The `stackTelemetry.enabled` setting will be `true` by default. Users will be
+notified during the Kubewarden stack install and upgrade via Helm NOTES.txt
+that this telemetry is included, and instructions to disable it will be
+provided
+
+## Telemetry Server
+
+The telemetry server will be an instance of the [Longhorn
+upgrade-responder](https://github.com/longhorn/upgrade-responder).
+This server will receive information from controllers and store it in an
```
jhkrug:
```suggestion
This server receives information from controllers and store it in an
```

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,413 @@
...
+- `city`: City name extracted from the request's source IP.
+- `country`: Country name extracted from the request's source IP.
+- `country_isocode`: Country ISO code extracted from the request's source IP.
+
+The same data point will have the following fields:
+
+- `policyServerCount`: The number of PolicyServers deployed in the cluster.
+- `policiesCount`: The number of policies deployed in the cluster.
+- `namespaceUid`: A UID to uniquely identify the Kubewarden stack installation.
+  This should be UID of the namespace where Kubewarden is installed.
```
jhkrug:
```suggestion
  This should be UID of the Kubewarden installation namespace.
```

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,413 @@
...
+  always set to `1`,
+  [added](https://github.com/longhorn/upgrade-responder/blob/a6f6c7736b7e420b07ae7d813765dac778ebc638/upgraderesponder/service.go#L560)
+  automatically by `upgrade-responder`. This default value ensures that the data
+  point has a value for metric calculations, even if the request contains no
+  other fields. In other words, it's a dummy InfluxDB field used to count the
+  number of data points
+
+  ### Request validation schema
+
+  upgrade-responder should be configured with a JSON schema used to validate the
```
jhkrug: Not sure about this one but how about:
```suggestion
  The `upgrade-responder` should be configured with a JSON schema used to validate the
```

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,413 @@
...
+
+The `upgrade-responder`, InfluxDB, and Grafana instances used for Kubewarden
+telemetry will be maintained by the SUSE team. All members of the Kubewarden
+organization on GitHub will have access to the collected data.
+
+# Drawbacks
+
+- When new metadata or more metric data is required, it necessitates code
+  changes and a new release for both the controller and, maybe, the telemetry
+  server. The JSON schema versions must be kept in sync.
```
jhkrug:
```suggestion
- When requiring new metadata or more metric data, code
  changes will be needed. Along with a new release for both the controller and, possibly, the telemetry
  server. The JSON schema versions must be kept synchronized.
```

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,413 @@
...
+
+# Alternatives
+
+The following subsection is a option given during the write of this RFC. But it
+was not select by the majority of the team. The team decided to move with the
+`upgrade-responder` option because this is a solution used by other CNCF
+projects (Longhorn) and we (SUSE Rancher - who is willing to pay for the
+infrastructure) also have experience managing and maintaining the server-side
+infrastructure
```
jhkrug:
```suggestion
The following subsection is an option provided during the writing of this RFC. But, it
was not selected by a majority of the team. The team decided to move forward with the
`upgrade-responder` option because this is a solution used by other CNCF
projects (Longhorn). Also, we (SUSE Rancher — who is willing to pay for the
infrastructure) have experience managing and maintaining the server-side
infrastructure.
```

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,413 @@
...
+
+## Not selected option: OpenTelemetry Integration
+
+This implementation option aims for a less intrusive approach, decoupling the
+controller from the telemetry backend. With this design, all telemetry will be
```
jhkrug:
```suggestion
controller from the telemetry back-end. With this design, all telemetry will be
```

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,413 @@
...
+managed outside the Kubewarden controller's reconciliation loop. The controller
+will be instrumented with telemetry collection points that send data to an
+OpenTelemetry collector, which will then be configured to forward this data to
```
jhkrug:
```suggestion
OpenTelemetry collector, which will be configured to forward this data to
```

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,413 @@
...
+    attribute.String("policy_server_image", policy.GetPolicyServer().GetImageDigest()),
+    attribute.String("module", policy.GetModuleDigest()),
+    attribute.String("kubewarden_version", KUBEWARDEN_VERSION),
+    attribute.String("instance_uid", getInstanceUID()), // Corrected key name for consistency
+}
+
+policiesGauge.Record(context.Background(), 1, metric.WithAttributes(commonAttrs...)) // Corrected context.Background()
+```
+
+Once the metric is recorded, it is up to the OpenTelemetry collector to
```
jhkrug:
```suggestion
Once the metric is recorded, it's up to the OpenTelemetry collector to
```

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,413 @@
...
+manipulate the data and send it to a remote location, which can be another
+OpenTelemetry collector or a time-series database (e.g., Prometheus).
+
+### OpenTelemetry Configuration
+
+In the OpenTelemetry collector configuration, a new pipeline should be added to
```
jhkrug:
```suggestion
In the OpenTelemetry collector configuration, a new pipeline will be added to
```

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,413 @@
...
+      exporters: [otlphttp/remoteKubewardenOtel] # Corrected to match exporter name
+```
+
+This OpenTelemetry configuration can be used to control the telemetry export
+and can be expanded over time to include more metadata by leveraging the rich
+ecosystem of OpenTelemetry
+[processors](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor),
+potentially with no changes to the controller code. This makes telemetry
+customization more flexible. In the example above, processors are used to add
+labels, filter unrelated metrics, and batch metrics before sending.
```
jhkrug:
```suggestion
This OpenTelemetry configuration can be used to control the telemetry export
and, be expanded over time to include more metadata by leveraging the rich
ecosystem of OpenTelemetry
[processors](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor).
This would potentially need no changes to the controller code. This makes telemetry
customization more flexible. In the example above, processors will be used to add
labels, filter unrelated metrics, and batch metrics before sending.
```

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,413 @@
...
+
+All OpenTelemetry collector configuration can be managed in the Helm chart
+installation/update. The remote endpoint would be provided in the Helm Chart
+values. If the `stackTelemetry.enabled` is disabled, the OpenTelemetry
+configuration will not be applied:
```
jhkrug:
```suggestion
configuration won't be applied:
```

rfc/0023-kubewarden-stack-telemetry.md
```diff
@@ -0,0 +1,413 @@
...
+  remoteOtelEndpoint: "https://metrics.kubewarden.io"
+  batchTimeout: "1h"
+```
+
+The remote OpenTelemetry collector's configuration is abstracted from the user.
+For instance, it could use the [GeoIP
+Processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/geoipprocessor)
+to add location information to the received metrics.
+
+# Unresolved Questions
```
jhkrug:
```suggestion
# Unresolved questions
```

jhkrug (changes requested):
Some small suggestions or improvements to consider. Thanks.

rfc/0023-kubewarden-stack-telemetry.md
viccuad: I still believe that the current wording is ambiguous. By "members of the Kubewarden org" do we mean those listed in https://github.com/orgs/kubewarden/people ? If so, I would link that there then.

Personally, I would be happier with "people with access to the cncf-kubewarden-maintainers at lists.cncf.io", which follows the approach of other projects we have reviewed. It may also help users in feeling that their data (even if it only contains metadata that is anonymous) is better guarded.

viccuad (commented):
The RFC is looking good to me; only missing a clarification on exactly who gets to access the telemetry metadata, for my approval.

rfc/0023-kubewarden-stack-telemetry.md
jvanz: The thing is that data that we send to `upgrade-responder` is the data that will land in the InfluxDB. Therefore, it affect the queries that we will able to perform and by consequence the metrics we are able to track. That's why I've shared detailed options. 

This means that if we want to rank policies the controller probably should fetch/manipulate more information by the policies which affect the changes described in this document. If we do not mentioned that, we could hide  required actions.  Therefore, I see two options:

1. Do not mentions policies metrics at all and keep that for a second iteration of the metrics
2. Mention the suggested metric to here and add a text later in the  controller changes stating something like "fetch all the required data to allow the proposed metrics" 

I choose option 2.

rfc/0023-kubewarden-stack-telemetry.md
jvanz: I've added your request in the document.

flavio (approved):
The RFC looks good to me, I'm fine merging it and submitting to CNCF for their approval

rfc/0023-kubewarden-stack-telemetry.md
viccuad: Many thanks! resolving.

viccuad (approved):
LGTM too!

robertsirc (approved):
LGTM, I like the direction of this.

PR #54, opened 2026-04-15
feat: Unified Kubewarden admission controller Helm chart

jvanz:
## Description

Adds a proposal of how to merge all the Kubewarden admissions controller Helm charts (kubewarden-crds, kubewarden-controller and kubewarden-defaults) into a single Helm chart.

This document is a compilation of the discussions from the issue about this desired of unification of the Helm charts.

### DIFF
--- a/rfc/0026-unified-admission-controller-chart.md
+++ b/rfc/0026-unified-admission-controller-chart.md
@@ -14,14 +14,13 @@ This RFC proposes merging the three existing Kubewarden Helm charts
 (`kubewarden-crds`, `kubewarden-controller`, and `kubewarden-defaults`) into a
 single admission-controller chart. CRDs are moved under `templates/crds` so
 they participate in the normal upgrade lifecycle. The `default` PolicyServer
-and the recommended policies are defined as full Kubernetes YAML in a ConfigMap
-that the controller reads periodically. A `DefaultsApplier` runs on a
-configurable ticker (like the existing `CertReconciler`); on each tick it
-decodes each resource into the corresponding typed CRD struct, injects
-ownership labels, and creates or updates it via the API server. Stale managed
-resources are cleaned up automatically. This simplifies the installation
-process from a multi-chart, order-dependent procedure to a single `helm
-install` command.
+and the recommended policies are defined as full Kubernetes YAML in a ConfigMap.
+A `DefaultsApplier` Reconciler watches the ConfigMap for changes; on each
+reconciliation it decodes each resource into the corresponding typed CRD
+struct, injects ownership labels, and creates or updates it via the API server.
+Stale managed resources are cleaned up automatically. This simplifies the
+installation process from a multi-chart, order-dependent procedure to a single
+`helm install` command.
 
 # Motivation
 
@@ -99,17 +98,20 @@ responsible for creating and maintaining them:
    The ConfigMap is a regular Helm-managed resource, so `helm upgrade`
    naturally updates it.
 
-2. **DefaultsApplier (periodic, leader-only)**: A `DefaultsApplier` component
-   is registered with the controller-runtime manager using the `Runnable` and
-   `LeaderElectionRunnable` interfaces (the same pattern used by the existing
-   `CertReconciler`). When the leader is elected, its `Start()` method begins a
-   periodic loop controlled by a configurable ticker. On each tick (and
-   immediately on startup) it runs the following phases:
+2. **DefaultsApplier (watch-based Reconciler, leader-only)**: A
+   `DefaultsApplier` is implemented as a controller-runtime reconciler. It
+   is registered with the manager using `ctrl.NewControllerManagedBy(mgr)` and
+   watches ConfigMap resources with a predicate filter that selects only the
+   target ConfigMap in the deployment namespace. The ConfigMap name defaults to
+   `kubewarden-defaults` and is configurable via the `--defaults-configmap-name`
+   CLI flag on the controller binary. The controller-runtime framework handles
+   leader election — the Reconciler only runs on the leader instance.
+   Reconciliation is triggered by ConfigMap create, update, and delete events.
+   On each reconciliation it runs the following phases:
 
    - **Phase 1 — Read ConfigMap**: Fetch the `kubewarden-defaults` ConfigMap
      from the deployment namespace. If the ConfigMap does not exist, clean up
-     all resources carrying the managed ownership labels and wait for the next
-     tick.
+     all resources carrying the managed ownership labels and return success.
    - **Phase 2 — Apply desired resources**: For each `*.yaml` key in the
      ConfigMap data, decode the value into the appropriate typed CRD struct
      (e.g., `v1.PolicyServer`, `v1.ClusterAdmissionPolicy`) using the
@@ -121,23 +123,22 @@ responsible for creating and maintaining them:
    - **Phase 3 — Clean up stale resources**: List all PolicyServers and
      ClusterAdmissionPolicies carrying the full set of ownership labels. Delete
      any that are absent from the desired set (the ConfigMap keys).
-   - **Wait**: Sleep until the next tick. The ticker duration is configurable
-     via a Helm value (e.g., `defaults.reconcileInterval`, default: `5m`). The
-     loop continues until the context is cancelled (controller shutdown).
 
    This ensures:
 
    - CRDs are expected to be registered before the controller starts (they are
      rendered under `templates/crds/`). If a CRD is not yet available, the
-     Create/Update call fails and the next tick retries automatically.
+     Create/Update call fails and controller-runtime requeues the reconciliation
+     automatically.
    - Recommended policies are cleanly removed when disabled via Helm values.
    - The managed `default` PolicyServer is removed when disabled, though this
      causes all policies bound to it to be deleted by the controller's existing
      reconciliation logic.
    - Idempotency is guaranteed by the Create-or-Update logic. Re-applying the
      same desired state is a no-op.
-   - On failure, the next tick retries the operation automatically.
-   - ConfigMap changes from `helm upgrade` are picked up on the next tick
+   - On failure, controller-runtime requeues the reconciliation with exponential
+     backoff.
+   - ConfigMap changes from `helm upgrade` trigger an immediate reconciliation
      without requiring a pod restart.
 
 ### ConfigMap structure
@@ -243,11 +244,15 @@ Documenting this clearly is important, though we do not mark the ConfigMap as
 
 ### DefaultsApplier implementation outline
 
-The `DefaultsApplier` follows the same pattern as the existing
-`CertReconciler`. It implements the `Runnable` and `LeaderElectionRunnable`
-interfaces from controller-runtime. The manager starts it only on the leader
-instance. It runs a periodic loop controlled by a configurable ticker,
-re-reading the ConfigMap and reconciling default resources on each tick.
+The `DefaultsApplier` is implemented as a controller-runtime reconciler. It
+is registered with the manager using `ctrl.NewControllerManagedBy(mgr)` and
+watches ConfigMap resources with a predicate filter that selects only the
+target ConfigMap by name. The ConfigMap name is provided via the
+`--defaults-configmap-name` CLI flag (default: `kubewarden-defaults`), which is
+exposed as a Helm value. The Helm chart passes this value to the controller
+Deployment's container args. Controller-runtime's
+built-in leader election ensures the Reconciler runs only on the leader
+instance.
 
 Because the ConfigMap holds full Kubernetes resource YAML and the controller
 already registers the Kubewarden CRD types in its scheme, each entry is decoded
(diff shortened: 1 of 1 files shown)


rfc/0026-unified-admission-controller-chart.md
```diff
@@ -0,0 +1,576 @@
...
+`values.yaml`. The `questions.yaml` for the Rancher UI is updated accordingly.
+The fact that a ConfigMap and Job are involved is hidden from the user.
+
+# Migration path
+[migration]: #migration
+
+This is the recommended approach for its simplicity. There is a window between
+uninstalling the old charts and completing the new installation where **no
+admission control is active** — any workload can be deployed unchecked during
+this period.
```
jvanz, line 370: Down time is not nice. I was looking for a solution to avoid this by making some magic labels updates. But decided to park this for now until we decided to move forward with this change.

rfc/0026-unified-admission-controller-chart.md
```diff
@@ -0,0 +1,576 @@
...
+  Workarounds: use `helm upgrade --dry-run` to preview changes, maintain
+  external backups, or implement deletion inventory tracking.
+- **Dual active policies during name changes**: because Phase 2a (apply)
+  runs before Phase 2b (cleanup), renaming a recommended policy between chart
+  versions creates a window where both the old-named and new-named policies are
+  simultaneously active on the cluster. If the two policies produce conflicting
+  admission decisions (one allows, one denies), observed behavior is
+  unpredictable for the duration of the Job run. Mitigation: avoid renaming
+  policies between versions; treat renames as a removal plus an addition and
+  document the brief dual-enforcement window in release notes.
```
jvanz: This list is  getting me anxious on thinking the complexity to keep this job running fine in multiple uninstall/install/reinstall/upgrade scenarios. Specially about the work to avoid orphans resources and possible security problems. How do you feel? This feels "wrong" to me.

rfc/0026-unified-admission-controller-chart.md
```diff
@@ -0,0 +1,576 @@
...
+upgraded like any other template resource.
+
+## Defining default resources via a ConfigMap + Job
+
+The `default` PolicyServer and the recommended admission policies are no longer
+rendered directly as Helm templates. Instead:
+
+1. **ConfigMap**: A ConfigMap is rendered by Helm containing:
+   - The full YAML definitions of the `default` PolicyServer and each
+     recommended policy, with Helm-templated ownership labels for safe cleanup.
```
flavio: Do we need to manually inject the helm ownership labels? I'm afraid this could back fire in some cases.

The cleanup you're talking about is when everything is removed. When that heppens, the CRDs are remove as well, leading to the removal of all the policies and policy servers.

rfc/0026-unified-admission-controller-chart.md
```diff
@@ -0,0 +1,576 @@
...
+   - A text inventory file listing desired default resource IDs,
+     used by the Job to identify stale resources during cleanup.
```
flavio: I'm not sure we need that

rfc/0026-unified-admission-controller-chart.md
```diff
@@ -0,0 +1,576 @@
...
+   volume and runs a two-phase algorithm:
+   - **Phase 1**: Wait for CRDs and controller readiness.
+   - **Phase 2a**: Iterate over each `*.yaml` file in the mounted ConfigMap path
+     and run `kubectl apply -f <file>` to create/update desired resources.
+     Non-YAML files (e.g., the `defaults-inventory.txt` inventory) are
+     explicitly excluded by the glob pattern.
+   - **Phase 2b**: Diff the cluster's managed default resources against the
+     desired inventory. Delete stale managed resources (those with the full set
+     of ownership labels but absent from the desired inventory), including the
+     `default` PolicyServer when it is disabled via Helm values.
```
flavio: This is not what I had in mind when I explained the whole idea. Let's discuss during the daily

rfc/0026-unified-admission-controller-chart.md
viccuad: > The cleanup you're talking about is when everything is removed. When that heppens, the CRDs are remove as well, leading to the removal of all the policies and policy servers.

I think the cleanup José is concerned about is the one on `helm upgrade`ing the charts and disabling options without removing CRDs. For example, disabling recommended policies.

This could be achieved by making the post-install/post-upgrade hook job be idempotent (remove all policies, then reinstate those enabled, etc). But that at point we would benefit from having a reconciling loop in the controller.

jvanz:
We (@kubewarden/kubewarden-developers ) discussed this RFC in our daily call. There is a communication issue from what I understood from the issue that originates this document. From the issue, I understood that we would like to avoid controller changes. However, we all agree now that the best approach is to use the controller to handle the creation of the Kubewarden default resources.  Therefore, I'll update the document to reflect that.

jvanz:
I'm moving this to the blocked column because I need to work in another [issue](https://github.com/kubewarden/kubewarden-controller/issues/1658) that we would like to add in the controller for the next release. I'll come back to this after that.

jvanz:
@kubewarden/kubewarden-developers I think you can review once more.

rfc/0026-unified-admission-controller-chart.md
```diff
@@ -0,0 +1,544 @@
...
+```
+
+**One ConfigMap key = one desired resource.** The set of `*.yaml` keys defines
+the desired state. The controller iterates over all keys ending in `.yaml`,
+decodes each into the appropriate typed CRD struct using the scheme, and
+creates or updates the resource via the API server.
+
+**Helm templates do the rendering.** All values (`{{ .Values.* }}`) are
+resolved at Helm template time. The controller receives fully-formed YAML and
+does not need to understand Helm values or custom configuration schemas.
```
flavio, line 207: I was initially skeptical about this approach, but I think it's a good compromise. I like it, since it keeps the boilerplate to the minimum inside of the Go codebase.

flavio (approved):
I like the design. I think, unrelated with this RFC, I'll create an issue to change how we currently handle the deletion of a Policy Server.

I think we should not delete all the policies that are scheduled on it. We should leave them pending.

Given this RFC is probably going to be managed as part of the 1.37 release (we just released 1.35), I think it would be nice to have this change land as part of 1.36.

Update: I've created https://github.com/kubewarden/adm-controller/issues/1700

rfc/0026-unified-admission-controller-chart.md
```diff
@@ -0,0 +1,544 @@
...
+# Detailed design
+
+[design]: #detailed-design
+
+## Chart restructuring
+
+The three existing charts are merged into a single chart
+(`kubewarden-admission-controller` or equivalent). CRDs are placed under
+`templates/crds/` (not the special `/crds` directory) so they are rendered and
+upgraded like any other template resource.
```
viccuad, line 80: Suggestion: rename the `kubewarden-controller` image to be `controller`.

flavio: As discussed on slack, let's do that right now as part of the 1.36 release

rfc/0026-unified-admission-controller-chart.md
```diff
@@ -0,0 +1,544 @@
...
+   at render time, producing complete resource manifests. Disabled resources
+   are simply omitted from the ConfigMap. When all defaults are disabled, the
+   ConfigMap is not rendered at all.
+
+   Helm values control which resources are included, their configuration (labels,
+   annotations, `namespaceSelector`, default registry, failure policy, etc.).
+   The ConfigMap is a regular Helm-managed resource, so `helm upgrade`
+   naturally updates it.
+
+2. **DefaultsApplier (periodic, leader-only)**: A `DefaultsApplier` component
```
viccuad: I'm ok with this approach, but I am curious on why we don't listen for changes in the ConfigMap (looking at its `generation` field instead of periodic polling). On failure, it would retry indefinitely. Is it because of a possible race condition with the CRDs being installed?

I am uneasy telling people "install the chart, but it's not enough with a `helm install --wait`, you will need to wait an arbitrary time  (~5 mins) for things to pop up". Feels opaque.

rfc/0026-unified-admission-controller-chart.md
```diff
@@ -0,0 +1,544 @@
...
+  validating and mutating webhooks, ensuring defaults are validated exactly like
+  user-created resources.
+- The controller overwrites managed resources entirely on each apply. User
+  customization of managed defaults must go through Helm values, not manual
+  edits.
+- Disabling the managed `default` PolicyServer is destructive because the
+  controller will delete all policies bound to it.
+- **Upgrade safety**: the ConfigMap YAML and the controller binary ship in the
+  same Helm chart, so the typed Go structs always match the ConfigMap schema.
+  During a rolling upgrade there is a brief window where the old binary reads
```
viccuad, line 319: I think it should be possible to mark the ConfigMap template as post-install/post-upgrade hook, meaning that it will get applied after the controller has been upgraded.

rfc/0026-unified-admission-controller-chart.md
```diff
@@ -0,0 +1,544 @@
...
+# Unresolved questions
+
+[unresolved]: #unresolved-questions
+
+- **PolicyServer deletion cascading to policies**: the current controller
+  behavior of deleting all policies when their PolicyServer is removed is
+  orthogonal to this RFC but amplifies the risk of any approach that deletes
+  the PolicyServer during upgrades. A separate investigation should evaluate
+  changing this to leave policies in a "pending" state instead.
+- **Orphaned managed resources handling**: should the cleanup logic detect orphaned
```
viccuad, line 546: My vote goes for a warning log.

viccuad (approved):
LGTM, approving.
Just a couple of musings. Also, given that we are tackling https://github.com/kubewarden/adm-controller/issues/1700, we could simplify the RFC and remove concerns throughout.

rfc/0026-unified-admission-controller-chart.md
jvanz: > I'm ok with this approach, but I am curious on why we don't listen for changes in the ConfigMap (looking at its generation field instead of periodic polling). On failure, it would retry indefinitely. Is it because of a possible race condition with the CRDs being installed?

No. There is no fundamental reason of using the periodic polling. Just following the partner we already use for the certificates and avoid the usage of verifications if the event we got in the watch is from the ConfigMap that we expect. 

>I am uneasy telling people "install the chart, but it's not enough with a helm install --wait, you will need to wait an arbitrary time (~5 mins) for things to pop up". Feels opaque.

This will not gonna happen. As stated in the document: `On each tick (and immediately on startup) it runs the following phases...`

If we decided to change to watch configmap changes, we need to add a initial trigger to ensure that the ConfigMap is applied in the beginning of the controller run.

rfc/0026-unified-admission-controller-chart.md
jvanz: I would leave that to the controller retry mechanism. I'm not big fan of hooks. Even if we use the hooks there is still the possibility, due the concurrency nature of the events when we have multiple replicas, that the controller will process a new version of the configmap before the upgrade finished.

rfc/0026-unified-admission-controller-chart.md
flavio: I think it would be better to react to `ConfigMap` updates. We've already a watch monitoring `ConfigMap` resources inside of the namespace where the controller is deployed (it syncs the PolicyServer config).

There's no need to write special code to handle the startup. The controller runtime will synthetically generate a `CREATE` even of the ConfigMap once the controller starts

rfc/0026-unified-admission-controller-chart.md
flavio: I don't think there's this problem. During the rollout the old instance of controller is turned down, the new one is ramped up. I think we will be good

rfc/0026-unified-admission-controller-chart.md
flavio: Let's see how it goes with the implementation, I don't think we could end up with resources losing their ownership labels...

rfc/0026-unified-admission-controller-chart.md
jvanz: I've updated the document to create a reconciler that will watch for the configmap change.

jvanz:
@flavio @viccuad I've update the document to document the creation of a reconciler that watches for configmap. I've marked some of your comments as resolved just to avoid noise in the discussions about this new approach. Please, open it again if you believe they are still valid. Thanks!

viccuad (approved):
Thanks!

rfc/0026-unified-admission-controller-chart.md
```diff
@@ -0,0 +1,560 @@
...
+`immutable` because Helm must be able to update it on `helm upgrade`.
+
+### DefaultsApplier implementation outline
+
+The `DefaultsApplier` is implemented as a controller-runtime reconciler. It
+is registered with the manager using `ctrl.NewControllerManagedBy(mgr)` and
+watches ConfigMap resources with a predicate filter that selects only the
+target ConfigMap by name. The ConfigMap name is provided via the
+`--defaults-configmap-name` CLI flag (default: `kubewarden-defaults`), which is
+exposed as a Helm value. The Helm chart passes this value to the controller
```
flavio, line 252: We can change that at implementation time, but I would not allow the user to pick the name of the configmap. We're going to have a hardcoded value.

flavio (approved):

